PR #182, opened 2015-03-27
Ros topic

barbalberto:
This pull request add 2 new features:
1- A wrapper can be instantiated on his own and attach to an already existing controlBoard, just by setting the wrapper configuration like:

joints 10
networks (left_arm left_hand)
left_arm      0 6  0 6
left_hand    0 2  0 2

The wrapper will search the singleton for two devices called left_arm and left_hand and, if found, attach to them.
If the devices are not found, it simply fails the open and it closes down. No magic here yet, this could be improved somehow in the future if needed. To use this, verify the wrapper-only is the last device in the sdf file.

2- If a [ROS] group parameter is found in the config file, it is propagated to the wrapper configuration. This is useful to enable controlBoardWrapper2 to dinamically create a ROS compatible topic. This feature is under testing on a yarp branch (checkout to yarp 'mergePortname_and_ROS' branch if you want to test it before it is released)

### DIFF
--- a/src/gazebo_plugins/ControlBoard.cc
+++ b/src/gazebo_plugins/ControlBoard.cc
@@ -19,7 +19,7 @@ namespace gazebo
 
 GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
 
-    GazeboYarpControlBoard::GazeboYarpControlBoard() : m_iWrap(0), m_controlBoard(NULL)
+    GazeboYarpControlBoard::GazeboYarpControlBoard() : m_iWrap(0)
     {}
 
     GazeboYarpControlBoard::~GazeboYarpControlBoard()
@@ -31,7 +31,8 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
         if (m_wrapper.isValid())
             m_wrapper.close();
 
-        GazeboYarpPlugins::Handler::getHandler()->removeDevice(m_driverName);
+        for (int n = 0; n < m_controlBoards.size(); n++)
+            GazeboYarpPlugins::Handler::getHandler()->removeDevice(m_controlBoards[n]->key.c_str());
 
         GazeboYarpPlugins::Handler::getHandler()->removeRobot(m_robotName);
         yarp::os::Network::fini();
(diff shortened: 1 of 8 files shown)


lornat75:
I think it could have been better to split this into two pull requests.

How is feature 1 novel? (I miss this :( )

barbalberto:
Actually the 1 is a middle step required in order to have the second one useful.

Right now gazebo plugins always create a pair made of a wrapper and its controlBoard device, like it happens when using the subdevice option.
It is not possible to create only a wrapper, neither to create only a controlBoard. This is because since it is a simulator, you can always configure the plugins to have a one-to-one match between wrapper and controlboard device, so there was no need to add more complexity.

traversaro:
Ok, two concerns: 
 * who is closing motorControl devices now?
 * In case of a wrapper connecting to multiple controlboards, are we correctly decrementing all the device used ?

barbalberto:
In case of multiple usage the decrement counter is correctly handled.
The close is done when counter reaches zero inside the removeDevice function.
In the previous commit, I forgot it while doing the changes, fixed in last commit

src/gazebo_plugins/ControlBoard.cc
```diff
@@ -109,41 +116,62 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
 
         for (int n = 0; n < netList->size(); n++)
         {
-            yarp::os::ConstString driverName( netList->get(n).asString().c_str());
+            m_driverName = netList->get(n).asString();
```
traversaro: On a wrapper handling multiple devices, only the last one is saved in `m_driverName` ?

barbalberto: Ops, yes.  I'll fix it. 
I didn't want to add too much complexity but it turns out to have more facets then I though at the beginning.

To add more variability, are we allowed in gazebo to have 2 or more iCub models loaded at the same time?
I think it'll be bettere to use robot scoped name ad a unique key to register devices to avoid duplicated name, for example 2 let_arm controlboards coming from different robots. 
What do you think?

traversaro: It make sense. An easy trick to avoid complicating code is creating a unique "scoped" device name as `deviceScopedName = robotScopedName+"::" + deviceName` .

traversaro:
@barbalberto Just to be sure that you are not waiting for my review, is this pull request ready or not?

barbalberto:
Sorry, I was busy with advr :( 
I made all the changes we discuss, I think it is ready now. If you are fine with it, I'll merge it.

I just merged the relevant changes in yarp into master branch so we can actually test it.

traversaro:
This PR broke support for running simulation with two robots that shared some part names (such as a coman and an iCub or a coman and a walkman) because the key used for the mapping in the singleton is not unique in the simulation.
cc @barbalberto @EnricoMingo

barbalberto:
Intresting ... I'll chek it next week

Issue #174, opened 2015-02-05
[Dissemination] Create homebrew formula


francesco-romano:
As specified in #95

francesco-romano:
A first, basic and maybe not working, implementation here https://github.com/robotology/gazebo-yarp-plugins/blob/3313b11c00149de0e400bf7485cad91463ab1287/utilities/installation/brew/gazebo-yarp-plugins.rb

francesco-romano:
using `yarp` as recommended dependency seems the best option:
by default yarp is built with brew. But it is possible to remove the dependency and specify by environment variable the `YARP_DIR` path.

Strange thing: in the brew environment I get this error:
~~~
bash-3.2$ cmake .
-- YARP is version: 2.3.63.5
-- Boost version: 1.57.0
CMake Error at /usr/local/lib/cmake/gazebo/gazebo-config.cmake:46 (find_package):
  By not providing "FindOGRE.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "OGRE", but
  CMake did not find one.

  Could not find a package configuration file provided by "OGRE" with any of
  the following names:

    OGREConfig.cmake
    ogre-config.cmake

  Add the installation prefix of "OGRE" to CMAKE_PREFIX_PATH or set
  "OGRE_DIR" to a directory containing one of the above files.  If "OGRE"
  provides a separate development package or SDK, be sure it has been
  installed.
Call Stack (most recent call first):
  CMakeLists.txt:16 (find_package)


-- Configuring incomplete, errors occurred!
See also "/tmp/gazebo-yarp-plugins-YaxywX/CMakeFiles/CMakeOutput.log".
~~~
but this not happens when building the project (cloned by github).
Any ideas @traversaro

francesco-romano:
@traversaro solved.
The `gazebo-config.cmake` uses `pkg-config`  and I had to add it as a dependency of the formula or in the "sandboxed" environment the command was not available.

francesco-romano:
Almost there. For some reasons the inclusion of OGRE is not passed to the compiler in the brew environment.

francesco-romano:
I just retried and cmake successfully configures the project, but I get
```bash
In file included from /tmp/gazebo-yarp-plugins20150423-51434-ygv2zs/src/gazebo_plugins/Camera.cc:8:
[...]

PR #186, opened 2015-04-23
Added Homebrew formula

francesco-romano:
At the current stage the formula is in an alpha stage, but the skeleton is ready.
The open issues are:

- only compiles with `--HEAD` option (the release is too old, but I'm leaving it as a placeholder).
- we should investigate if we can install also the icub-gazebo models in the formula
- tests can (should) be enabled
- double `cmake` command to make it work :disappointed: 
- have a clean `audit`
```bash
brew audit gazebo-yarp-plugins 
gazebo-yarp-plugins:
 * Use the `cp` Ruby method instead of `system "cp"`

Error: 1 problems in 1 formulae
```

related issue: https://github.com/robotology/gazebo-yarp-plugins/issues/174

@traversaro


traversaro:
We should revert my changes in .travis.yml because gazebo5 is not available on precise (12.04).

francesco-romano:
@traversaro done

Issue #187, opened 2015-04-30
Build system misc issues


traversaro:
We recently experienced with @francesco-romano  that the file structure in this repo is quite confusing, especially for a new user that want to create/edit a plugin. 

We were then thinking to migrate to a structure where every plugin has its own directory with source code and every other resource (thrift) required. This should simplify a lot the structure of the repo. 

Given that there are no open PR, and given that the `addObjects` branch would anyway need a serious manual intervention to merge in master, I think now is a good time to proceed with this change. 

Any opinions ? 

cc @barbalberto @EnricoMingo @arocchi  @MirkoFerrati @jeljaik @prashanthr05

barbalberto:
Fine by me. If the split between gazebo_plugin and yarp_driver folders will be removed is a plus for me.

traversaro:
Exactly!

jeljaik:
totally agree

--


*Jorhabib Eljaik Gómez,**Ph.D. Student - *Istituto Italiano di Tecnologia
Dept. of Robotics, Brain and Cognitive Sciences
Cognitive Humanoids Laboratory
e-mail: jorhabib.eljaik@iit.it
Cel: +39 388 352 33 85
Tel: +39 010 71781 420
Fax: +39 010 7170817

MirkoFerrati:
Fine. But don't think that addObjects requires that much effort to be
merged! The main effort is due to change of filenames and variable names!
I will merge in June if it's fine for you :)

traversaro:
Yes, for the problems of merging 'addObjects' I was referring to the workaround committed for the clock issue.

francesco-romano:
There are still some issues
- [ ] `yarp_idl`: when it autogenerates the files it does not "respect" the namespace convention: a header file with namespace `gazebo` is not saved in filesystem as `include\gazebo\file.h`. This creates issues then in the inclusion. Maybe we can open an issue in yarp
- [ ] I started to use `target_include_directories` instead of exporting global variables. The only question here is: what is the priority for the headers? build or install dir?
- [ ] I created a separate library for the thrift (currently only the clock, but the logic remains the same): this should allow clients to properly link to the autogenerated library instead of copy-paste sources (see https://github.com/robotology-playground/WBI-Toolbox for the `ySynchronizer` block). But probably I'm missing the `config.cmake` file.
- [ ] Tests should be enabled again (but anyway what is the status of the tests?)

francesco-romano:
[...]

PR #188, opened 2015-05-02
New repository structure.

francesco-romano:
- libraries: contains project-wise libraries (which must be built before any plugin)
- thrift: contains rpc targets. Each target should build to a separate library (with relative includes) so that it can be linked by clients
- plugins: contains the plugins

Fixes https://github.com/robotology/gazebo-yarp-plugins/issues/187


francesco-romano:
cc @traversaro for a first check.

There is still some issues (which may be solved after merge, or not)
- [ ] `yarp_idl`: when it autogenerates the files it does not "respect" the namespace convention: a header file with namespace `gazebo` is not saved in filesystem as `include\gazebo\file.h`. This creates issues then in the inclusion.
- [ ] I started to use `target_include_directories` instead of exporting global variables. The only question here is: what is the priority for the headers? build or install dir?
- [ ] I created a separate library for the thrift (currently only the clock, but the logic remains the same): this should allow clients to properly link to the autogenerated library instead of copy-paste sources (see https://github.com/robotology-playground/WBI-Toolbox for the `ySynchronizer` block). But probably I'm missing the `config.cmake` file.

Any other suggestions are welcome :smile:

francesco-romano:
Unfortunately I bumped the cmake required version (not explicitly yet) to 3.0.0

libraries/common/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+# CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
+
+cmake_minimum_required(VERSION 2.8.11)
+
+PROJECT(Library_Common)
+
+add_library(gazebo_yarp_lib_common INTERFACE)
+target_include_directories(gazebo_yarp_lib_common INTERFACE
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+)
```
traversaro: Cool, I did not know you could use this trick for header only libraries.

traversaro:
What is requiring CMake 3.0 ? This could be a problem on Ubuntu 12.04 and Debian 7.

francesco-romano:
The trick I used for the header only library. I discovered this only after

francesco-romano:
@traversaro I added the possibility to specify a plugin as a module (for now default is shared).
I removed the requirement on CMake 3.0

traversaro:
I think that thee issues in https://github.com/robotology/gazebo-yarp-plugins/pull/188#issuecomment-98341893 can be solved also after the merge. 
The blocking problem that I see at this moment is that we are breaking the documentation. because is not possible anymore to add the build directory to the `GAZEBO_PLUGIN_PATH` environmental variable, because all the plugins are now scattered in several subdirectories. I guess this is ok, but we should update the documentation and send a warning to the codyco users that the installation procedure has changed.

francesco-romano:
@traversaro I updated the documentation accordingly

traversaro:
Perfect, then we can merge. :+1: 

Public announcement : 
@azadm @robertocalandra @barbalberto @EnricoMingo @MirkoFerrati @rlober @hu-yue @mikearagao 
Please notice that after this change, it is not really feasible to use the plugin directly from the build directory. Check the updated documentation for how to properly use the plugins from the install directory:
https://github.com/francesco-romano/gazebo-yarp-plugins/commit/51d052196b45a9c6efc9d168956ebdeaee6c3412
(Please propagate this information to your colleagues , thanks!)

PR #194, opened 2015-05-18
Enable plugin testing on travis file 

traversaro:
Using [RTF](https://github.com/robotology/robot-testing) and [icub-tests](https://github.com/robotology/icub-tests). 

Currently the tests are running the https://github.com/robotology/icub-tests/blob/master/suits/basics-icubGazeboSim.xml suite, but this can be changed and we can eventually provide a suite directly in the `gazebo-yarp-plugins` repository. 

Curiously, the gazebo fixture does not work unless we start (and then kill) an instance of gzserver before launching the server (this should be properly investigated).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ before_script:
   - cd icub-tests
   - mkdir build
   - cd build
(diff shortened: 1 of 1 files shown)


traversaro:
Closing in favor of https://github.com/robotology/gazebo-yarp-plugins/pull/233 .

PR #196, opened 2015-07-02
[imu, camera] Update Stamp count only when sending data

drdanz:
I'm not sure if this is correct, but I think that the stamp count should be updated only when the data is sent by yarp, and not when it is generated by gazebo...
That makes it easier to count if any data was missed because of YARP

Implemented only for imu and camera plugins, I don't know what the other plugins do.

### DIFF
--- a/plugins/camera/include/yarp/dev/CameraDriver.h
+++ b/plugins/camera/include/yarp/dev/CameraDriver.h
@@ -107,7 +107,8 @@ private:
     bool m_display_time_box;
     bool m_display_timestamp;
 
-    yarp::os::Stamp m_lastTimestamp; //buffer for last timestamp data
+    double m_lastTimestamp;          //buffer for last timestamp data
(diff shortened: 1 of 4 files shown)


traversaro:
I understand the specific use case, but this is rather inconsistent with respect to the behavior of the camera and the imu device on the iCub:
* https://github.com/robotology/icub-main/blob/master/src/libraries/icubmod/dragonfly2/linux/FirewireCameraDC1394-DR2_2.h#L74
* https://github.com/robotology/icub-main/blob/master/src/libraries/icubmod/xsensmtx/XSensMTx.cpp#L327

Unless the devices in the robot (I checked only iCub, but I guess that coman will be implemented in a similar way) are modified to follow this new behavior, I would prefer to maintain a consistency between the simulator and the real robot.

drdanz:
Ok no problem! Closing this.

PR #201, opened 2015-09-03
Worldinterface2

lornat75:
This PR implements a plugin for controlling the world in gazebo. At the moment it adds methods to create simple objects like sphere, box and cylinder. Important TODOs are:

* Revise interface after some usage, possible improvements could be to split functions in simpler ones to reduce the number of parameters to pass from command line (although not really important)
* Check performance in synchronous and asynchronous mode
* Implement loading of models from file
* Add example client code
* Make poses relative to the robot

Known issue: removing and adding models probably causes a leak due to how Gazebo implements removal of models.


traversaro:
I am starting reviewing the PR, if you prefer we can move it to an internal branch so everyone can modify it.

traversaro:
PR moved to https://github.com/robotology/gazebo-yarp-plugins/pull/202 .

PR #202, opened 2015-09-07
Worldinterface2

traversaro:
Original pull request by @lornat75 : 
~~~
This PR implements a plugin for controlling the world in gazebo. At the moment it adds methods to create simple objects like sphere, box and cylinder. Important TODOs are:

    Revise interface after some usage, possible improvements could be to split functions in simpler ones to reduce the number of parameters to pass from command line (although not really important)
    Check performance in synchronous and asynchronous mode
    Implement loading of models from file
    Add example client code
    Make poses relative to the robot

Known issue: removing and adding models probably causes a leak due to how Gazebo implements removal of models.
~~~


traversaro:
First modification: disable compilation of the worldinterface for Gazebo < 4.0 . I still prefer to mantain compatibility of the basic plugins with Gazebo 2.2 because it is used in ROS Indigo, that is a ROS LTS Release (EOL in April, 2019). However I think it is fine to offer the world interface only for users with Gazebo >= 4.0 .

plugins/worldinterface/src/worldproxy.cpp
```diff
@@ -0,0 +1,348 @@
...
+{
+  sdf::SDF sphereSDF;
+ 
+  string sphereSDF_string=string(
+      "<?xml version='1.0'?>\
+       <sdf version ='1.4'>\
+          <model name ='sphere'>\
+            <pose>POSEX POSEY POSEZ ROLL PITCH YAW</pose>\
+            <link name ='link'>\
+              <pose>POSEX POSEY POSEZ ROLL PITCH YAW</pose>\
```
traversaro: Are you sure about this duplicated pose ? This will basically mean that the model frame is placed with the requested pose in the world, and then the link frame is placed with the requested pose wrt the model pose. In the end you applied the requested pose two times, I think it make sense to just put the requested pose in the model tag.

plugins/worldinterface/src/worldproxy.h
```diff
@@ -0,0 +1,193 @@
+// -*- mode:C++; tab-width:4; c-basic-offset:4; indent-tabs-mode:nil -*-
+/*
+ * Copyright (C) 2015 iCub Facility 
+ * Authors: Lorenzo Natale
+ * CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
+ *
+ */
+
+#ifndef __WORLDPROXY__
+#define __WORLDPROXY__
```
traversaro: A modification just for being picky: the C++ standard reserves the use of identifier starting with a double underscore:
~~~
— Each identifier that contains a double underscore __ or begins with an underscore followed by an uppercase letter is reserved to the implementation for any use.
— Each identifier that begins with an underscore is reserved to the implementation for use as a name in the global namespace.
~~~
I'll modify it to fit what we use in the other files.

plugins/worldinterface/src/worldproxy.cpp
lornat75: Not sure, I took it verbatim from http://gazebosim.org/tutorials?tut=plugins_world; initially I found it strange, then I completely forgot about it...

lornat75:
I also forgot to mention there are some compilation warnings which I meant to revise but did not have time to.

traversaro:
The plugin is not affecting any existing functionality, so we can merge it and push any new improvement directly in the master branch.

PR #203, opened 2015-09-29
[controlboard] implemented function to set/get position pids

traversaro:
Implemented the setPid/getPid for setting position pids. 

cc @francesco-romano 

Two question for @randaz81 : 
 * The IPidControl methods just have influence on position gains, regardless of the controlmode in which the controlboard is right? 
 * I have implemented them such that the unit of measurement of the gains are degrees-based for revolute joints because I remember that with EMS-based robots we will migrate to "metric" gains.. I am forgetting right or I am doing something wrong?

### DIFF
--- a/libraries/common/include/GazeboYarpPlugins/common.h
+++ b/libraries/common/include/GazeboYarpPlugins/common.h
@@ -32,6 +32,20 @@ namespace GazeboYarpPlugins {
     double convertRadiansToDegrees(double radians);
 
     /**
+     * \brief convert a degrees-based gain to a radians-based gain
+     * \param gainsWrtDegrees degrees-based gain
+     * \return the gain converted in radians-based units
+     */
+    double convertDegreesGainsToRadiansGains(double gainsWrtDegrees);
+
+    /**
+     * \brief convert a radians-based gain to a degrees-based gain
+     * \param gainsWrtRadians radians-based gain
+     * \return the gain converted in degrees-based units
+     */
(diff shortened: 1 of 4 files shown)


traversaro:
Still todo : 
  * [ ] set/get all other kind of pids/gains 
  * [ ] add loading of gains in configuration files in the new format, bet keeping backward compatibility

libraries/common/include/GazeboYarpPlugins/common.h
```diff
@@ -52,6 +66,17 @@ namespace GazeboYarpPlugins {
             return false;
         }
     }
+
+    inline double convertDegreesGainsToRadiansGains(double gainsWrtDegrees)
+    {
+        return gainsWrtDegrees * 180.0 / pi;
```
francesco-romano: Are you sure the implementations are not switched?

francesco-romano:
@traversaro  a part from the comment above, I think we can still merge and move the pending tasks to an issue. What do you think?

libraries/common/include/GazeboYarpPlugins/common.h
traversaro: Let's dothe math for proportional gains:
[gainsWrtDegrees] = [outputQuantity] / deg 
[gainsWrtRadians] = [outputQuantity] / rad 
so gainsWrtRadians = gainsWrtDegrees * deg/rad = gainsWrtDegrees * 180.0 / pi
But if you need to ask it clearly need a better explanation in the documentaiton.

PR #204, opened 2015-10-12
Add support for using gazebo-provided configuration parameters in yarp configuration files 

traversaro:
To simplify passing using Gazebo values (such as the runtime robot
name or sensor name) in yarp configuration files, now some special
yarp configuration values (for now gazeboYarpPluginsRobotName and
gazeboYarpPluginsSensorName) are present when loading the plugins
configuration file. 

In this way the user can either ignore them,
either use them using the ${variableName} syntax, either overwriting
them specifying explicit value for them in the configuration files.

To understand the use of this feature, take the icub model from 
https://github.com/robotology-playground/icub-gazebo/tree/master/icub 
and comment out the `gazeboYarpPluginsRobotName icubGazeboSim` in
 https://github.com/robotology-playground/icub-gazebo/blob/master/icub/conf/gazebo_icub_robotname.ini . In this way the gazeboYarpPluginsRobotName 
configuration value is not overwritten by the user configuration files, and so its value
(and hence the prefix of all the ports opened by the robot)  will be the one of the robot in Gazebo .
Given that any robot has a unique name in Gazebo, using this feature is possible to 
spawn multiple robots from the same model without yarp port names conflicts. 

Whitespace-free diff available at https://github.com/robotology/gazebo-yarp-plugins/pull/204/files . 

cc @EnricoMingo

### DIFF
--- a/libraries/singleton/CMakeLists.txt
+++ b/libraries/singleton/CMakeLists.txt
@@ -14,6 +14,7 @@ add_gazebo_yarp_plugin_target(LIBRARY_NAME singleton
                               SYSTEM_INCLUDE_DIRS ${YARP_INCLUDE_DIRS} ${Boost_INCLUDE_DIRS} ${GAZEBO_INCLUDE_DIRS} ${SDFORMAT_INCLUDE_DIRS} ${PROTOBUF_INCLUDE_DIRS}
                               LINKED_LIBRARIES ${YARP_LIBRARIES} ${SDFORMAT_LIBRARIES} ${GAZEBO_LIBRARIES} ${PROTOBUF_LIBRARIES} ${Boost_LIBRARIES}
                               HEADERS include/GazeboYarpPlugins/Handler.hh
+                                      include/GazeboYarpPlugins/ConfHelpers.hh
                               SOURCES src/Handler.cc
-                                 )
+                                      src/ConfHelpers.cc)
 
--- a/libraries/singleton/include/GazeboYarpPlugins/ConfHelpers.hh
+++ b/libraries/singleton/include/GazeboYarpPlugins/ConfHelpers.hh
@@ -1,0 +1,101 @@
+/*
+ * Copyright (C) 2007-2013 Istituto Italiano di Tecnologia ADVR & iCub Facility
+ * Authors: Enrico Mingo, Alessio Rocchi, Mirko Ferrati, Silvio Traversaro and Alessandro Settimi
+ * CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
+ */
+
+#ifndef GAZEBOYARP_CONFIGURATION_HELPERS_HH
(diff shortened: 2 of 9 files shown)


barbalberto:
I'm not sure to understand ... 
To activate this feature I have to export the variable like
 ```"export gazeboYarpPluginsRobotName=pippo"```   #first comment, usually our env var are all capital with words separated by underlines

I see that when both are defined, the one environment will prevail over the config file, but if neither of the two is defined the robot prefix name is missing and ports names start with a ```//```

In my case, including 2 iCubs in gazebo still raise a port conflict because both try to use either the config file name or the same env variable. Can you explain better what is the desired outcome for 2 instance if the robot?

Some other questions:

Removing and re-inserting the same robot will end up with different port names?
Does it avoid name conflicts also for rosTopic and rosNode names when activated?

traversaro:
Probably "environmental variable" is a confusing name. : ) 
This are just yarp config values that are "prefiled" by gazebo when loading the configuration files, not actual "enviromental variables". In this way, you can refer to them when writing yarp ini configuration files, using the `${confParamName}` syntax.

barbalberto:
Ok, so if I comment out the parameter  ```gazeboYarpPluginsRobotName  icubGazeboSim ``` from the config file, then the robotname will be the one choose by Gazebo simulator, am I right?

traversaro:
Exactly. Regarding you questions: 

> Removing and re-inserting the same robot will end up with different port names?

Yes, but this is gazebo bug ( https://bitbucket.org/osrf/gazebo/issues/1142/potential-for-smart-pointer-reference ). Once that bug is solved I guess it would be possible to removing and re-inserting and keep the same name.

> Does it avoid name conflicts also for rosTopic and rosNode names when activated?
If you use the configuration parameters set by the plugins in the configuration file, yes.

traversaro:
(I somehow put some C++11 specific code in the PR, I will correct it in the near future (I hope).

barbalberto:
Ok, now I've understood, thanks.
However it seems that on a first test, on my pc doesn't work. If I comment out the param from the config file, it simply starts without it (no robotname in the ports)

barbalberto:
Ops, sorry, it does work!! It was my mistake.

EnricoMingo:
Do we merge? I try it and it works.

traversaro:
We need to fix the [`std::vector<std::string> splitString(const std::string &s, const std::string &delims)`](https://github.com/robotology/gazebo-yarp-plugins/pull/204/files#diff-6374fb85025a47ba5644e9484e15b2e7R27) to avoid using C++11, otherwise we will break compilation of users of Gazebo < 4 . I can work on it this evening, if anyone is able to fix it is doing me a favor.

traversaro:
Pushed a C++98 compatible version.

traversaro:
@barbalberto @EnricoMingo any further comments?

barbalberto:
I'm loving it!! :+1:

traversaro:
:+1:

PR #207, opened 2015-10-30
Fix compilation with Gazebo 7

traversaro:
In particular :
  * Address function that are migrating to ignition-math classes
  * Address migration of SDF from boost::shared_ptr to std::shared_ptr

### DIFF
--- a/libraries/singleton/include/GazeboYarpPlugins/ConfHelpers.hh
+++ b/libraries/singleton/include/GazeboYarpPlugins/ConfHelpers.hh
@@ -7,9 +7,13 @@
 #ifndef GAZEBOYARP_CONFIGURATION_HELPERS_HH
 #define GAZEBOYARP_CONFIGURATION_HELPERS_HH
 
-#include <yarp/os/Semaphore.h>
+#include <sdf/sdf_config.h>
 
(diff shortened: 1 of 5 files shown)


traversaro:
cc @francesco-romano @barbalberto

libraries/singleton/include/GazeboYarpPlugins/ConfHelpers.hh
```diff
@@ -22,11 +25,6 @@ namespace gazebo
     }
 }
 
-namespace sdf {
-    class Element;
-    typedef boost::shared_ptr<Element> ElementPtr;
-}
```
traversaro: `ElementPtr` is : 
~~~
typedef boost::shared_ptr<Element> ElementPtr;
~~~
in Gazebo < 7 and 
~~~
typedef std::shared_ptr<Element> ElementPtr;
~~~
In Gazebo >= 7 . 
I think it is easier and more readable to just include `#include <sdf/Element.hh>` instead of having a forward declaration with a lot of `#ifdef`.

francesco-romano: :+1:

traversaro:
Travis finally passing, merging.

PR #216, opened 2015-11-13
Implemented (set/get)TorquePid(s) methods

francesco-romano:
These methods do nothing from the control point of view. They simplify the test of modules which needed them.
These values are just stored internally during the set and retrieved during the get.


traversaro:
:+1:

Issue #192, opened 2015-05-13
SEA simulation using gazebo-yarp-plugins 


traversaro:

traversaro:
cc @hu-yue

traversaro:
We did some preliminary testing with @hu-yue and it seems that the SEA implementation proposed in discussed in the previous post actually works. 

I will then do a brief recap of which modification to [gazebo_yarp_controlboard](https://github.com/robotology/gazebo-yarp-plugins/tree/master/plugins/controlboard) are then needed to match the SEA behavior in real robots (i.e. iCub v2.5 legs).

### Load configurations for SEA joints
* [ ] Currently `gazebo_yarp_controlboard` is used to implement a simple controller running on the top of a 1DOF Gazebo joint. It loads the controlled gazebo joint [in the `jointNames` variable of the configuration file](https://github.com/robotology/gazebo-yarp-plugins/blob/6bdb463c5d049559ba2b4551f0e5e8ade273490e/plugins/controlboard/src/ControlBoardDriver.cpp#L217), as in https://github.com/robotology-playground/icub-gazebo/blob/master/icub/conf/gazebo_icub_left_leg.ini : 
~~~
# Specify configuration of MotorControl devices
[left_leg]
# name of the device to be instatiated by the factory
device gazebo_controlboard
#jointNames list
jointNames l_hip_pitch l_hip_roll l_hip_yaw l_knee l_ankle_pitch l_ankle_roll  
name left_leg
~~~
We can think of an alternative way of specifying the jointNames, such that is possible to specify more complex options such as elastic joints, as in:
~~~
# Specify configuration of MotorControl devices
[left_leg]
# name of the device to be instatiated by the factory
device gazebo_controlboard
#jointNames list
jointNamesAdvanced ( ( (actuatedJoint l_hip_pitch) (type normal) ) ( (actuatedJoint l_hip_roll) (type normal) ) ( (actuatedJoint l_hip_yaw) (type normal) ) ( (actuatedJoint l_knee) (elasticJoint l_knee_elastic) (type elastic) ) ( (actuatedJoint l_ankle_pitch) (elasticJoint l_ankle_pitch_elastic) (type elastic) ) ( (actuatedJoint l_ankle_roll) (elasticJoint l_ankle_roll_elastic) (type elastic) ) )
name left_leg
~~~ 
[...]

Issue #165, opened 2014-12-29
Stop using Gazebo JointController in the ControlBoard plugin


traversaro:
And directly uset setForce()/addForce methods in the controlboard plugin.

Currently the low level position and velocity PIDs are updated at every Gazebo [WorldUpdateBegin](https://github.com/robotology/gazebo-yarp-plugins/blob/master/src/yarp_drivers/ControlBoardDriver.cpp#L89) event by sending references to the Gazebo JointController through the Gazebo protobuf internal middleware (see `sendPositionToGazebo`, `sendVelocityToGazebo`and `sendTorqueToGazebo` methods). This messages are then received by the Gazebo internal [JointController](https://bitbucket.org/osrf/gazebo/src/7d2654784d9e7fba7088e6837155f4982fb6a3d3/gazebo/physics/JointController.cc?at=default#cl-154). This messages are then used in the `JointController::Update()` method, that is itself called by the main gazebo physics thread update method, and use the `Joint::SetForce` method to actually apply torque to the joints. 
I always found this disturbing for two reasons:
- We basically have a function (our plugin `onUpdate`) that is called my the physics thread, that sends some data to the physics thread itself  through an external middleware. This is both ugly and inefficient. 
- We are bound to use internal Gazebo PID structure as a black box, while for example for supporting Mixed Position-Velocity control or Coupled DOFs we would prefer to have full control of the control loop. 

For this reason, the first step could be to move the control logic completely in the `gazebo_yarp_controlboard`, by calling the `SetForce` in the `OnUpdate` method. This can be simplified by using the same `gazebo::common::PID` object already used in the JointController. 
Once we have complete control of the control loop, we can easily program any kind of behaviour actually implemented in the low level firmware, such as coupled mechanism control.

Discussing starting from here: https://github.com/robotology/gazebo-yarp-plugins/commit/5c2173f0362a1c598ba958baddaefe032955d834 , the rationale behind this issue was discussed here: 
https://github.com/robotology/gazebo-yarp-plugins/issues/149 .

mikearagao:
Hi @traversaro, I'm back on this, do you know if someone already tried to run my last commit with the iCub?

Cheers!

traversaro:
Ok @mikearagao , if you think it should work as master we can test it!

mikearagao:
I will try to run it with the iCub before merge. I will try to do this later on the afternoon, I'll keep you updated!

traversaro:
[...]

Issue #149, opened 2014-11-03
iCub/Vizzy eyes/Gaze support


traversaro:
In this issue we will discuss more in detail what it is needed for implementing a simulation of iCub eyes controlled by the iKinGazeCtrl module. 

## Eye URDF/SDF model : 
At IIT we recently started to set-up a pipeline for automatic extraction of  robot models directly from Creo CAD models, using [SimMechanics Link](http://www.mathworks.it/help/physmod/smlink/index.html) and [simmechanics-to-urdf](https://github.com/robotology-playground/simmechanics-to-urdf). Using this pipeline, we can investigate the extraction of the head/eye kinematic structure directly from CAD.

## Camera simulation support in `gazebo-yarp-plugins`:
Most of the work has been already done by @barbalberto in issue https://github.com/robotology/gazebo-yarp-plugins/issues/96 . The plugin still needs some polishing for supporting all the necessary functionality, but the basic support is already there. 

## Necessary enhancements to `gazebo_yarp_controlboard` plugin.
Currently, only the 3DOF neck mechanism is simulated in Gazebo. The eyes have some peculiarities, for which some modification to the controlboard plugin will be necessary.

### Controlled DOFs
Disregarding details on the actuators placements (as we do in the rest of the Gazebo simulation)
the eye form a quite standard 3DOF tree structure: the first DOF is `eyes_tilt` i.e. the common tilt of the two eyes, while the other 2 DOFs are the two independent "pans" of the two eyes, `left_eye_pan` and `right_eye_pan`. 

You can see this mechanism in this image of the first version of the iCub eyes:  
![eyes](https://cloud.githubusercontent.com/assets/1857049/4891494/829ea8a8-63a8-11e4-9206-81d78fb85de8.png)

However, the yarp controlboard is not controlling independently the two pans, but it is controlling the `eyes_version` and `eyes_vergence`, that are coupled DOFs related to the pans by the following equation (check for reference: http://wiki.icub.org/wiki/Vergence,_Version_and_Disparity):
```
eyes_version    =  left_eye_pan + right_eye_pan
eyes_vergence   =  left_eye_pan - right_eye_pan
```
[...]

PR #217, opened 2015-11-16
References once & trajectory generation

barbalberto:
Thi PR aims to fix a problem on trajectories and improve the way references are sent to Gazebo internal simulation core in the following ways:

- If there is no new reference does not publish new ref to Gazebo. This is required for
  CER robot because 2 plugins will use the same Gazebo Joint and otherwise they will
  always conflict onto each other

- Previous trajectory generator had the following problem: if vel is very high and threshold to stop the   generator very small it could happend that the minimum increment step is higher than the threshold used
to  stop the trajectory. In this case it'll oscillate around the final position but never converge.

- Also uses new way of setting position references to set initial configuration

### DIFF
--- a/plugins/controlboard/src/ControlBoardDriver.cpp
+++ b/plugins/controlboard/src/ControlBoardDriver.cpp
@@ -20,7 +20,8 @@ using namespace yarp::sig;
 using namespace yarp::dev;
 
 
-const double RobotPositionTolerance = 0.9;
+const double RobotPositionTolerance_revolute = 0.9;      // Degrees
+const double RobotPositionTolerance_linear   = 0.004;    // Meters
 
 GazeboYarpControlBoardDriver::GazeboYarpControlBoardDriver() : deviceName("") {}
 GazeboYarpControlBoardDriver::~GazeboYarpControlBoardDriver() {}
@@ -46,6 +47,7 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
     m_torques.resize(m_numberOfJoints); m_torques.zero();
     m_trajectoryGenerationReferenceSpeed.resize(m_numberOfJoints);
     m_referencePositions.resize(m_numberOfJoints);
+    m_oldReferencePositions.resize(m_numberOfJoints);
     m_trajectoryGenerationReferencePosition.resize(m_numberOfJoints);
     m_trajectoryGenerationReferenceAcceleraton.resize(m_numberOfJoints);
     m_referenceTorques.resize(m_numberOfJoints);
@@ -59,13 +61,9 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
     m_minDamping.resize(m_numberOfJoints, 0.0);
     m_maxDamping.resize(m_numberOfJoints, 100.0);
     m_jointTypes.resize(m_numberOfJoints);
+    m_positionThreshold.resize(m_numberOfJoints);
 
-    if(!configureJointType() )
-        return false;
-
-    setMinMaxPos();
-    setMinMaxImpedance();
-    setPIDs();
+    // Initial zeroing of all vectors
     m_positions.zero();
     m_zeroPosition.zero();
     m_referenceVelocities.zero();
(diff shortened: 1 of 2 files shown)


plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -404,9 +433,15 @@ bool GazeboYarpControlBoardDriver::sendPositionsToGazebo(Vector &refs)
 bool GazeboYarpControlBoardDriver::sendPositionToGazebo(int j,double ref)
 {
     gazebo::msgs::JointCmd j_cmd;
-    prepareJointMsg(j_cmd,j,ref);
-    m_jointCommandPublisher->WaitForConnection();
-    m_jointCommandPublisher->Publish(j_cmd);
+
+    if(ref != m_oldReferencePositions[j])
+    {
+        // std::cout << "Sending new command: new ref is " << ref << "; old ref was " << m_oldReferencePositions[j] << std::endl;
```
traversaro: Debug print to clean?

plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -115,16 +135,11 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
...
-        for (unsigned int i = 0; i < m_numberOfJoints; ++i) {
-#if GAZEBO_MAJOR_VERSION >= 4
-            m_jointPointers[i]->SetPosition(0,initial_config[i]);
-#else
-            gazebo::math::Angle a;
-            a.SetFromRadian(initial_config[i]);
-            m_jointPointers[i]->SetAngle(0,a);
-#endif
-        }
+        // Set initial reference
```
traversaro: This will be changing the behavior of the initial configuration tag from spawning the robot in the initial configuration, to just give the initial configuration as the initial desired position to the low level PID. This will result in a initial abrupt motion for models in wich the initial configuration is not the zero position (for example iCub arms). 


There is any reason for this change? @barbalberto

traversaro:
I don't know the details of how the CER plugin will work, but consider that there is a longstanding effort to get rid of the use of the Gazebo JointController in the ControlBoard plugin, see https://github.com/robotology/gazebo-yarp-plugins/issues/165 and https://github.com/robotology/gazebo-yarp-plugins/issues/192 . I hope this two use case don't conflict.

plugins/controlboard/src/ControlBoardDriver.cpp
barbalberto: It is commented out so no noisy prints. I just kept in case the need of debugging should arise.

traversaro: I understand that this is subjective, but I find that if we keep all the old debug messages in the code the code will become quickly a unreadable mess of commented out prints... 

However given that this is just my personal taste, no problem!

barbalberto:
Thanks for the comment. This PR shouldn't change *how* the references are sent to the Gazebo internals, but only *how often*. 
The 'constraint' is that if no new reference are received from ports (or trajectory is ended) everything stays in place without the need of setting the same reference over and over again ( for position and position direct mode).
The needs come from the fact that 2 different 'motionControl' plugins needs to control the same Gazebo joint.

plugins/controlboard/src/ControlBoardDriver.cpp
barbalberto: I see what you means, thank you for this clarification.
The reason was to use a uniform way to set the initial value taking care of rotationale and linear joints. I'll find a better way to handle it.

traversaro:
The problem is that to control coupled joints (see https://github.com/robotology/gazebo-yarp-plugins/issues/149 , https://github.com/robotology/gazebo-yarp-plugins/issues/165 and https://github.com/robotology/gazebo-yarp-plugins/issues/192  for discussion on this) we will need to move the low level PID simulation from the Gazebo JointController to the Gazebo ControlBoard plugin . 

As far I can see, this is incompatible to the notion of having two motionControl plugin controlling the same joint, unless an explicitly switch between the motionControl plugins is planned. What happens if the two motionControl try to control the same joint at the same time?

traversaro:
The easiest solution perhaps is to support both way of controlling a joint, either using the Gazebo JointController or using a PID inside the plugin, and specify in the configuration file which one to use. This will increase complexity but should cover both use cases.

barbalberto:
>  What happens if the two motionControl try to control the same joint at the same time?
This of course will cause a conflict, but it is the same as having 2 user applications, like MotorGui, to control the same joint. This cannot be avoided I think.

The point here is that the 'master' version of plugin is setting the a reference to Gazebo internals on each update, even if no new ref is actually sent from user. This means that one of the two is shadowed by the other.

If you plan to stop using Gazebo PID control, btw I like the idea, which will be the 'entry point' into Gazebo? I mean, from issues you referenced, my understanding is that we compute the PID and then directly set the torque applied on the joint? Is it right?

traversaro:
> This of course will cause a conflict, but it is the same as having 2 user applications, like MotorGui, to control the same joint. This cannot be avoided I think.
> The point here is that the 'master' version of plugin is setting the a reference to Gazebo internals on each update, even if no new ref is actually sent from user. This means that one of the two is shadowed by the other.

Ok, I see. 

> If you plan to stop using Gazebo PID control, btw I like the idea, which will be the 'entry point' into Gazebo? I mean, from issues you referenced, my understanding is that we compute the PID and then directly set the torque applied on the joint? Is it right?

Exactly.

barbalberto:
Ok, I understand.

With CER we need to call a iKin library between user inputs and Gazebo simulation, so this is a very specific functionality and it cannot be embedded into standard plugin.

Maybe the only long term solution could be to have the CER plugin became a wrapper of the standard one. I thought at the beginning to try this way but I faced other problems which I don't exactly recall now.

However, the sum is: yes, those changes do conflicts each other.
Proposal: as long as we do not get rid of Gazebo Internal PID, we can accept this change, it'll do no harm. (as soon as I've fixed the initialization part)

When you guys starts working on the new way and wants to merge into master we can adopt the temporary solution to use a parameter to choose between Gazebo/Custom PIDs as you suggest.

In the long run, the CER plugin will became a wrapper of standard motion control if possible and correct from a simulation perspective (I've to discuss about this with other people involved in CER project).

I'm quite interested in this idea of change the PID simulation engine, I heard about some change due to SEA but I admit I didn't fully understood the implications. 
I'd like to be in the loop of this topic because if we change this maybe we can make it more general and leverage on 'external' libraries.

Do you think creating a Wiki page with a small 'roadmap' explaining the planned work would be useful? I think the issue tends to be too dispersive when dealing with planning future work, but it's just my opinion.

traversaro:
A roadmap document would definitely be a good idea. In the meantime the solution contained in this PR will definitely work fine.

barbalberto:
:+1:
I'll fix the initialization and cleanup a bit the debug messages and then I think I can merge.
Thank you for this discussion, it helped me get a better understanding.

traversaro:
:+1: for merging.

PR #222, opened 2016-01-12
Fix warnings in Gazebo 6.5 and compilation in Gazebo 7.0

traversaro:
Cleanup also trailing whitespaces.


traversaro:
Version without whitespaces changes: https://github.com/robotology/gazebo-yarp-plugins/pull/222?w=1 .

Issue #229, opened 2016-02-23
Controlboard regression


traversaro:
@naveenoid  have reported some strange behaviours (joint reaching their limits on startup) with the new minimum jerk trajectory generator. While we investigate this we can switch back the default trajectory generator to the constant velocity one.

naveenoid:
:+1:

traversaro:
Everything seems to work fine at https://github.com/robotology/gazebo-yarp-plugins/commit/605f4292b47751dfac46dfd1dfb943876798e1b9 .

randaz81:
Issue fixed in master (missing initial speed value?)

PR #230, opened 2016-02-23
[controlboard] use constant_speed by default

traversaro:
See https://github.com/robotology/gazebo-yarp-plugins/issues/229 .

@naveenoid can you check with this branch?


traversaro:
This does not solve https://github.com/robotology/gazebo-yarp-plugins/issues/229 , closing.

PR #232, opened 2016-03-16
Fix gazebo7

barbalberto:
Update plugins to Gazebo 7:

- Fix crash of depthCamera on Gazebo7.
- Substitute all deprecated functions with new version.

### DIFF
--- a/plugins/depthCamera/include/yarp/dev/DepthCameraDriver.h
+++ b/plugins/depthCamera/include/yarp/dev/DepthCameraDriver.h
@@ -267,9 +267,10 @@ private:
     unsigned char  *m_RGBPointCloud_Buffer;
     int counter;
 
+#if GAZEBO_MAJOR_VERSION < 7
     gazebo::sensors::CameraSensor*       m_imageCameraSensorPtr;
     gazebo::rendering::CameraPtr         m_imageCameraPtr;
-
+#endif
     gazebo::sensors::DepthCameraSensor*  m_depthCameraSensorPtr;
     gazebo::rendering::DepthCameraPtr    m_depthCameraPtr;
 
--- a/plugins/depthCamera/src/DepthCameraDriver.cpp
+++ b/plugins/depthCamera/src/DepthCameraDriver.cpp
@@ -188,22 +188,21 @@ bool GazeboYarpDepthCameraDriver::OnNewImageFrame(const unsigned char *_image,
 //                     "\n\t_depth is " << _depth     <<
 //                     "\n\t_format is " << _format   << std::endl;
 
-
-#if GAZEBO_MAJOR_VERSION >= 7
     m_colorFrameMutex.wait();
+
+    #if GAZEBO_MAJOR_VERSION >= 7
     if(m_depthCameraSensorPtr->IsActive())
         memcpy(m_imageFrame_Buffer, m_depthCameraPtr->ImageData(), m_imageFrame_BufferSize);
 
     m_colorTimestamp.update(this->m_depthCameraSensorPtr->LastUpdateTime().Double());
-    m_colorFrameMutex.post();
 #else
-    m_colorFrameMutex.wait();
     if(m_imageCameraSensorPtr->IsActive())
         memcpy(m_imageFrame_Buffer, m_imageCameraSensorPtr->GetImageData(), m_imageFrame_BufferSize);
 
     m_colorTimestamp.update(this->m_imageCameraSensorPtr->GetLastUpdateTime().Double());
-    m_colorFrameMutex.post();
 #endif
+
+    m_colorFrameMutex.post();
     return true;
 }
 
--- a/plugins/depthCamera/src/DepthCameraDriver.cpp
+++ b/plugins/depthCamera/src/DepthCameraDriver.cpp
@@ -27,8 +27,10 @@ GazeboYarpDepthCameraDriver::GazeboYarpDepthCameraDriver()
     m_display_time_box  = false;
     m_display_timestamp = false;
 
+#if GAZEBO_MAJOR_VERSION < 7
     m_imageCameraSensorPtr = 0;
     m_imageCameraPtr       = 0;
+#endif
     m_depthCameraSensorPtr = 0;
     m_depthCameraPtr       = 0;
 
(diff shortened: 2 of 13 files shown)


plugins/depthCamera/src/DepthCameraDriver.cpp
```diff
@@ -161,19 +170,28 @@ bool GazeboYarpDepthCameraDriver::OnNewImageFrame(const unsigned char *_image,
...
+
+#if GAZEBO_MAJOR_VERSION >= 7
+    m_colorFrameMutex.wait();
+    if(m_depthCameraSensorPtr->IsActive())
+        memcpy(m_imageFrame_Buffer, m_depthCameraPtr->ImageData(), m_imageFrame_BufferSize);
+
+    m_colorTimestamp.update(this->m_depthCameraSensorPtr->LastUpdateTime().Double());
+    m_colorFrameMutex.post();
+#else
     m_colorFrameMutex.wait();
```
traversaro: This mutex wait/post can go outside the #if blocks (it is shared between the two branches).

plugins/depthCamera/src/DepthCameraDriver.cpp
```diff
@@ -120,23 +127,25 @@ bool GazeboYarpDepthCameraDriver::close()
...
     {
-        m_imageCameraPtr->DisconnectNewImageFrame(this->m_updateImageFrame_Connection);
+        m_depthCameraPtr->DisconnectNewImageFrame(this->m_updateImageFrame_Connection);
         this->m_updateImageFrame_Connection = gazebo::event::ConnectionPtr();
     }
 
     if (this->m_updateRGBPointCloud_Connection.get())
     {
-        m_imageCameraPtr->DisconnectNewImageFrame(this->m_updateRGBPointCloud_Connection);
+        m_depthCameraPtr->DisconnectNewImageFrame(this->m_updateRGBPointCloud_Connection);
```
traversaro: Is this `m_updateRGBPointCloud_Connection` member used anywhere in the code?

plugins/depthCamera/src/DepthCameraDriver.cpp
```diff
@@ -97,7 +104,7 @@ bool GazeboYarpDepthCameraDriver::open(yarp::os::Searchable &config)
 //     memset(m_RGBPointCloud_Buffer, 0x00, m_RGBPointCloud_BufferSize);
```
traversaro: The `m_RGBPointCloud_Buffer` seems not to be used at all, but is anyway deleted in https://github.com/robotology/gazebo-yarp-plugins/pull/232/files#diff-eff3bb8f5948742479396b9546d89fecR159 . Can we remove it?

plugins/depthCamera/src/DepthCameraDriver.cpp
barbalberto: Actually the usage is commented out, because Gazebo does not produce pointCloud yet, only rgb and depth.
The code is there in case in future update of Gazebo they'll implement this feature we are ready.

plugins/depthCamera/src/DepthCameraDriver.cpp
traversaro: Same for OnNewRGBPointCloud method.

plugins/depthCamera/src/DepthCameraDriver.cpp
```diff
@@ -120,23 +127,25 @@ bool GazeboYarpDepthCameraDriver::close()
     // Copy & paste like a Monkey
```
traversaro: Can we remove the debug prints some at line 115?

plugins/depthCamera/include/yarp/dev/DepthCameraDriver.h
```diff
@@ -267,9 +267,10 @@ class yarp::dev::GazeboYarpDepthCameraDriver:   public yarp::dev::DeviceDriver,
     unsigned char  *m_RGBPointCloud_Buffer;
     int counter;
 
+#if GAZEBO_MAJOR_VERSION < 7
     gazebo::sensors::CameraSensor*       m_imageCameraSensorPtr;
     gazebo::rendering::CameraPtr         m_imageCameraPtr;
```
traversaro: This pointers does not seem to be used at all. Can we remove them?

barbalberto: They are not used from gazebo 7 onwards, while still used for gazebo 6

traversaro: I see, I was missing this line : https://github.com/robotology/gazebo-yarp-plugins/blob/fix_gazebo7/plugins/depthCamera/src/DepthCameraDriver.cpp#L190 .

barbalberto:
Implemented hints and cleanup from comments

plugins/depthCamera/src/DepthCameraDriver.cpp
traversaro: I see, the method is already in the DepthSensor in Gazebo (see https://osrf-distributions.s3.amazonaws.com/gazebo/api/dev/classgazebo_1_1rendering_1_1DepthCamera.html#ab637c12f03aa1dc407d737a70b182ce5) but is not working? O_o

traversaro:
Great! Can be merged for me.

plugins/depthCamera/src/DepthCameraDriver.cpp
barbalberto: Exactly. The interface has the method, but it throws only garbage ...  I tried.
Also ROS plugins does not use it and compute the pointCloud manually. You can see here:
http://docs.ros.org/jade/api/gazebo_plugins/html/gazebo__ros__openni__kinect_8cpp_source.html#l00184

traversaro: Related issue:
* https://bitbucket.org/osrf/gazebo/issues/1865/rendering-depthcamera-does-not-output
* http://answers.gazebosim.org/question/5469/does-gazebo_ros_depth_camera-with-points-output-produce-a-color-point-cloud/

PR #233, opened 2016-03-25
Enable RTF-based icub-tests on travis

traversaro:


traversaro:
Finally working! 

For now only a very limited subset of functionalities are being tested (just the fact that the force torque ports stream correctly) and this will need several refinements, among them: 
* Encode testing logic as a target in `icub-tests` (so you can simply download `icub-tests`, run cmake and run `make test` to run the gazebo tests, without having to hardcode a lot of logic in the ci scripts). 
* Fix most test in `icub-tests` to run on Gazebo: currently most tests are not running for different reason (for example the test was assuming some hardware-specific behaviour or gazebo-yarp-plugins is not correctly implementing some interfaces). 

However is still nice to test the fact that the plugin are correctly running on each PR.

PR #234, opened 2016-04-01
Add examples for camera, depthcamera and laser

drdanz:


tutorial/model/camera/camera.ini
```diff
@@ -0,0 +1,5 @@
+device grabber
+subdevice gazebo_camera
+name /camera
+freq 30
```
barbalberto: What is `freq`? Shouldn't it be `framerate`?

traversaro:
:+1:
Ideally we could link them from here : https://github.com/robotology/gazebo-yarp-plugins/wiki/Embed-gazebo-yarp-plugins-in-an-SDF-model , but also having them in the tutorial directory is already a good step!

tutorial/model/camera/camera.ini
drdanz: Good catch, I copied it from the iCub model, I have no idea why it's using `freq` there, since it is not working... Should be fixed now

traversaro:
Thanks!

PR #235, opened 2016-04-08
Add multicamera plugin

drdanz:
Allows to add multicameras containing any number of cameras.
The cameras are supposed to be sync, but there is still some race condition that sometimes (on my pc ~1 or 2 frames each second) outputs 2 images with a different timestamp. Anyway it works good enought to be used.

### DIFF
--- a/plugins/multicamera/include/gazebo/MultiCamera.hh
+++ b/plugins/multicamera/include/gazebo/MultiCamera.hh
@@ -1,0 +1,62 @@
+/*
+ * Copyright (C) 2013-2015 Fondazione Istituto Italiano di Tecnologia RBCS & iCub Facility & ADVR
+ * Authors: see AUTHORS file.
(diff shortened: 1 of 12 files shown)


plugins/multicamera/include/gazebo/MultiCamera.hh
```diff
@@ -0,0 +1,62 @@
...
+#include <yarp/dev/PolyDriver.h>
+
+#include <yarp/dev/FrameGrabberInterfaces.h>
+
+#include <string>
+
+namespace gazebo
+{
+    namespace sensors {
+        class MultiMultiCameraSensor;
```
traversaro: Typo?

drdanz: More likely a double replacement `Camera` -> `MultiCamera` :smiley:

traversaro:
For me it can be merged.

PR #241, opened 2016-05-23
Maissensor

randaz81:


barbalberto:
I'll suggest to rename the sensor to a more meaningful name.
I suppose MaisSensor is supposed to emulate what the mais does on the real robot, but very few people know what a mais actually is (and what it does).
IMHO it is too tailored for the icub-like robots, while it will be better to name it based on the functionality implemented by the device, so that people from outside our lab can understand and use it.

Can the yDebug() be removed or changed into something like yTrace() to pollute less the output?

randaz81:
I promise to do it after the merge

Issue #243, opened 2016-05-26
New YARP workflow 


traversaro:
Historically, the `gazebo-yarp-plugins` master branch has been compatible with the master branch of `YARP` (in the past we try to be compatible also with the latest YARP release, but due to the breaking change in YARP at some point we just gave up). 

With the new workflow of YARP, this policy should be updated. 
For example this commit: https://github.com/robotology/gazebo-yarp-plugins/commit/4ac18d614abbee3e7cc49735f90fb979088eeabd broke compilation of `gazebo-yarp-plugins`, because it depended on a breaking change present only in the `devel` branch of https://github.com/robotology/yarp/commit/2159b37e1d984ba6f0794f61b41cc19014a4adba . 

For the time being (to avoid breaking the compilation of `gazebo-yarp-plugins` for most users) I created a `devel` branch also in `gazebo-yarp-plugins`, containing the commit that depends on the `devel` branch in yarp, while reverting the same commit in the `master` branch. 
However a proper policy, discussing for example if the master of `gyp` be compatible with YARP `master` and `devel`, should be prepared. 
If the coupling between `gazebo-yarp-plugins` and YARP interfaces is indeed so high, a possible convenient option is to just move `gazebo-yarp-plugins` in the `yarp` repository.

traversaro:
@randaz81 @drdanz @lornat75 @aerydna

barbalberto:
yes, you are right. 
master of `gazebo-yarp-plugins` should be compatible with master of `yarp`.
That was my fault, I've not thought this commit would break compilation.

traversaro:
Yes, that was a combination of factor, including the fact that according to the new workflow, no method should be removed from public API : https://github.com/robotology/yarp/blob/master/.github/CONTRIBUTING.md#deprecated-features . 
I don't know if it may be worth to document in the workflow document eventual exception for this role (for example for interfaces that are still a work in progress).

drdanz:
Sorry, my fault.
I didn't notice that @aerydna's commit changed the src/libYARP_dev/include/yarp/dev/IDepthSensor.h file as well, I thought it was changing only the driver. We can restore the old method, marking that as deprecated, or maybe change it back, since yarp::os::Property derives from yarp::os::Searchable... I don't know if the change to Property is really that necessary... @barbalberto what do you think?

barbalberto:
The change is necessary, and that interface is gonna change a lot in the next future since it is a wip.
[...]

PR #249, opened 2016-07-01
Clean up robot in destructor of GazeboYarpMaisSensor

Tobias-Fischer:
This lead to issues if the iCub is removed and added again, as the robot was still present.

### DIFF
--- a/plugins/maissensor/src/MaisSensor.cc
+++ b/plugins/maissensor/src/MaisSensor.cc
@@ -38,6 +38,7 @@ GazeboYarpMaisSensor::~GazeboYarpMaisSensor()
         m_wrapper.close();
 
     GazeboYarpPlugins::Handler::getHandler()->removeDevice(m_sensorName);
+    GazeboYarpPlugins::Handler::getHandler()->removeRobot(m_robotName);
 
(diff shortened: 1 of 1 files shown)


Tobias-Fischer:
@traversaro Would be great if you could merge.

traversaro:
I was outside of the lab today, I will check it tomorrow, thanks for the contribution!

traversaro:
As in https://github.com/robotology/gazebo-yarp-plugins/pull/250 : 
> Great catch, I think this could explain some strange error that we had in the past that were difficult to debug.

> As you may have read in https://github.com/robotology/gazebo-yarp-plugins/issues/243 , this repository switched to the new workflow in a rather unexpected way , so we did not fully document the expected workflow (sorry!). However I think it make sense to push bug-fixes such as this one in the `master` branch, and then forward merge them to the `devel` branch as described in https://github.com/robotology/yarp/blob/master/.github/CONTRIBUTING.md#example . 

> If you could retarget this PR against `master` it would be great, then I will forward merge it to `devel` as soon as it is merged. Let me know if you don't have the time to retarget the PR, in this case I will take care of it.

PR #250, opened 2016-07-01
Remove unnecessary m_robotRefreshPeriod

Tobias-Fischer:
This lead to a segfault in some cases as `GetPhysicsEngine()` pointed to `NULL` and there was no check against this.

### DIFF
--- a/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
+++ b/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
@@ -391,7 +391,6 @@ private:
 
     yarp::os::Property m_pluginParameters; /**< Contains the parameters of the device contained in the yarpConfigurationFile .ini file */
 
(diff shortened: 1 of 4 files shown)


Tobias-Fischer:
@traversaro would be great if you could merge :).

traversaro:
Great catch, I think this could explain some strange error that we had in the past that were difficult to debug.

As you may have read in https://github.com/robotology/gazebo-yarp-plugins/issues/243 , this repository switched to the new workflow in a rather unexpected way , so we did not fully document the expected workflow (sorry!). However I think it make sense to push bug-fixes such as this one in the `master` branch, and then forward merge them to the `devel` branch as described in https://github.com/robotology/yarp/blob/master/.github/CONTRIBUTING.md#example . 

If you could retarget this PR against `master` it would be great, then I will forward merge it to `devel` as soon as it is merged. Let me know if you don't have the time to retarget the PR, in this case I will take care of it.

PR #251, opened 2016-07-02
[depthSensor] Disable compilation of depthsensor in Gazebo 6

traversaro:
Travis on the `devel` branch has been failing for some time due to the fact that the depthSensor plugin does not compile under Gazebo6 (see https://travis-ci.org/robotology/gazebo-yarp-plugins/builds/139712221). 
Given that as far as I know nobody is using that sensor with Gazebo <= 6, I think it make sense to just drop support for that sensor in Gazebo 6.


traversaro:
Travis works fine, and the depth plugin is compiling fine on Gazebo 7  (see https://travis-ci.org/robotology/gazebo-yarp-plugins/jobs/141865477#L5254), merging to fix travis. 

cc @aerydna

PR #252, opened 2016-07-04
Clean up robot in destructor of GazeboYarpMaisSensor

Tobias-Fischer:
See https://github.com/robotology/gazebo-yarp-plugins/pull/249


traversaro:
Thanks a lot!

PR #253, opened 2016-07-04
Remove unnecessary m_robotRefreshPeriod

Tobias-Fischer:
See https://github.com/robotology/gazebo-yarp-plugins/pull/250


Tobias-Fischer:
@traversaro: Yes, you're right, it should be merged into `master` and from there in `devel`. Would be great if you could do that :).

Thanks, Tobi

traversaro:
Thanks a lot!

PR #255, opened 2016-07-15
Ros analog wrapper

barbalberto:
Propagate ROS parameter from .ini to analogWrapper in order to use ROS topic from inside gazebo


traversaro:
👍

PR #263, opened 2016-12-01
Control board warnings

Yeshasvitvs:
ControlBoard [DEBUG] and [INFO] output for stiffness and damping parameters suppressed

see https://github.com/robotology/icub-gazebo/issues/24 @traversaro


traversaro:
@randaz81 Just to check with you: I think this print need to be disabled in normal code.

traversaro:
Thanks!

PR #264, opened 2016-12-07
Depth camera update

aerydna:
DepthCamera plugin updated according to the new yarp's RGBDInterface (2.3.68)


traversaro:
@aerydna The tests were failing due to the fact that the port prefix of the icub gazebo simulator was already changed in `icubSim` some days ago. I fixed `icub-tests` to use the new port prefix.

traversaro:
Thanks!

PR #265, opened 2016-12-07
Revert "Control board warnings"

randaz81:
Reverts robotology/gazebo-yarp-plugins#263


traversaro:
@randaz81 There is a way to disable this debug prints?

Issue #154, opened 2014-11-13
[controlboard] implement missing methods in GazeboYarpControlBoardDriver


traversaro:
Several methods are not implemented in GazeboYarpControlBoardDriver.
- [x] `relativeMove(const int n_joint, const int *joints, const double *deltas)`
- [x] `checkMotionDone(const int n_joint, const int *joints, bool *flags)`
- [x] `setRefSpeeds(const int n_joint, const int *joints, const double *spds)`
- [x] `getRefSpeeds(const int n_joint, const int *joints, double *spds)`
- [x] `getRefAccelerations(const int n_joint, const int *joints, double *accs)`
- [x] `stop(const int n_joint, const int *joints)`
- [x] `setPositionDirectMode()`
- [x] `getEncoderAcceleration(int j, double *spds)`
- [x] `getEncoderAccelerations(double *accs)`
- [x] All the methods of `IPidControl` interface

If we want to use `gazebo_yarp_controlboard` for unit test of the controlboard related-software (simulated controlboard, real controlboard, wrappers) we should implement them. 

cc @EnricoMingo @arocchi

traversaro:
The rest of methods are returning false as of https://github.com/robotology/gazebo-yarp-plugins/issues/267 and are not implemented in other `icub-main`controlboards. I consider this issue to be closed when we merge https://github.com/robotology/gazebo-yarp-plugins/issues/267 .

traversaro:
Closed as https://github.com/robotology/gazebo-yarp-plugins/issues/279 was closed.

PR #267, opened 2016-12-20
[controlboard] missing methods implementation in GazeboYarpControlBoardDriver

Yeshasvitvs:
See https://github.com/robotology/gazebo-yarp-plugins/issues/154
@traversaro Thank you
In ControlBoardDriverPidControl.cpp, the following methods are not implemented. The semantics of this methods are not defined in the YARP interfaces, and they are not implemented in any known implementation of the YARP Controlboards
1) getErrorLimit
2)getErrorLimits
3)resetPid
4)disablePid
5)enablePid
6)setOffset

### DIFF
--- a/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
+++ b/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
@@ -403,6 +403,7 @@ private:
 
     yarp::sig::Vector m_positions; /**< joint positions [Degrees] */
     yarp::sig::Vector m_velocities; /**< joint velocities [Degrees/Seconds] */
+    yarp::sig::Vector m_accelerations; /**< joint accelerations [Degrees/Seconds/Seconds]> */
     yarp::sig::Vector m_torques; /**< joint torques [Netwon Meters] */
     yarp::sig::Vector m_maxTorques; /**< joint torques [Netwon Meters] */
 
--- a/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
+++ b/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
@@ -403,7 +403,7 @@ private:
 
     yarp::sig::Vector m_positions; /**< joint positions [Degrees] */
     yarp::sig::Vector m_velocities; /**< joint velocities [Degrees/Seconds] */
-    yarp::sig::Vector m_accelerations; /**< joint accelerations [Degrees/Seconds/Seconds]> */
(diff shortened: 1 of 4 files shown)


plugins/controlboard/src/ControlBoardDriverPidControl.cpp
```diff
@@ -97,8 +97,22 @@ namespace yarp {
...
+                *limit = 0.0;
+                return false;
+            }
+        }
+        
+        bool GazeboYarpControlBoardDriver::getErrorLimits (double *limits) {
+            if(!limits) return false;
+            for (unsigned int i=0; i<m_numberOfJoints; i++){
+                getErrorLimit(i,&limits[i]);
+            }
```
traversaro: A `return false` is missing here.

Yeshasvitvs: @traversaro my bad! fixed it

plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
```diff
@@ -403,6 +403,7 @@ class yarp::dev::GazeboYarpControlBoardDriver:
 
     yarp::sig::Vector m_positions; /**< joint positions [Degrees] */
     yarp::sig::Vector m_velocities; /**< joint velocities [Degrees/Seconds] */
+    yarp::sig::Vector m_accelerations; /**< joint accelerations [Degrees/Seconds/Seconds]> */
```
traversaro: I don't know the precedence rules between the `/` operator, but I guess that `Degrees/Seconds^2` would be less ambigous!

barbalberto:
+1 for implementing new stuff.

> These methods are also not implemented in YARP

What do you mean with that? YARP is middleware just moving data around. If an implementationhas to be found it is in iCub's devices. Can you explain better please?

traversaro:
> > These methods are also not implemented in YARP

> What do you mean with that? 

This should be substituted by "the semantics of this method is not defined in the YARP interfaces, and they are not implemented in any known implementation of the YARP Controlboards".

plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
Yeshasvitvs: @traversaro modified the units format

aerydna:
this pr start to smell funny.. @traversaro or @Yeshasvitvs  can you check if it's still valid, resolve the conflicts so we can merge it?

traversaro:
I would just close it until the semantics for those methods is defined in YARP.

Issue #226, opened 2016-02-07
Cease to support Gazebo 2.2 , 3 and 4 ?


traversaro:
Since last January, Gazebo 2.2 , 3 and 4 are not supported anymore by the OSRF (see http://gazebosim.org/#collapseVersion1_9 ). 

Given that making sure that the `gazebo-yarp-plugins` are compiling against all the old version of Gazebo is quite time consuming (it involves add a lot of `#ifdef` for whatever function changed signatures across versions) it may be convenient to drop support for older version of Gazebo.
Furthermore, older version of Gazebo has several well known issues https://github.com/robotology/gazebo-yarp-plugins#old-gazebo-versions so dropping support for them is a way of making sure that users use the newer versions of Gazebo. 

The current users I am aware of are: 
- iCub users are currently all using the latest version of Gazebo. 
- Walkman/Coman users are using currently using by default the Gazebo5 version with ROS Indigo (see http://gazebosim.org/tutorials?tut=ros_wrapper_versions) 
- Vizzy documentation (in https://github.com/vislab-tecnico-lisboa/vizzy#environment) claim that they try to be compatible with the default Gazebo versions in Hydro and Indigo (i.e. 1.9 and 2.2). In that case, dropping support for Gazebo 2.2 would break compilation for them. @plinioMoreno can you confirm that you have this requirement?

francesco-romano:
:+1: 

Or, if for Vizzy old gazebo versions are a must, we can branch out the code, and support only (critical) bug fixes to the old versions of the plugins (still time consuming, but less time consuming)

traversaro:
The problem of that option is that once in a while there are also modifications to `gazebo-yarp-plugins` to make sure that it still compiles fine against the latest yarp, and we would need to backport also those changes to the `vizzy` branch. If this is the case, until a proper release process is deployed for yarp I think that the least time consuming option is still to support old Gazebo version in the same codebase.

plinioMoreno:
Sorry for the late reply. About Vizzy, the gazebo 2 is just because of the simulation of the mobile platform and it is not related to the motor control of gazebo-yarp-plugins. In fact in the tests we did for the version-vergence controller we were using gazebo 5, because earlier versions do not have the Joint interfaces (external pointers) accessible outside of gazebo.

So from Vizzy side you can go ahead and drop support on the older gazebo versions, which is not needed for gazebo-yarp-plugins.

traversaro:
[...]

Issue #61, opened 2014-02-15
Check Gazebo version at compile time


traversaro:
Currently gazebo_yarp_plugins require at least Gazebo version 2.0 and Yarp 2.3.60, but the Gazebo requirement is not enforced in the build system.

Probably need a change in Gazebo upstream to support exporting of version numbers in CMake configuration files.

traversaro:
Given that version info is exported in Gazebo CMake config files since 4.0 [1] and currently we are only supporting Gazebo >= 5.0 [2] we can finally implement this.

[1] : https://bitbucket.org/osrf/gazebo/pull-requests/1153/generate-a-cmake-version-file-issue-1226/diff
[2] : loadConfigSensorPlugin

traversaro:
Fixed by https://github.com/robotology/gazebo-yarp-plugins/pull/268 .

PR #268, opened 2016-12-20
CMake Gazebo version check at compile time

Yeshasvitvs:
see https://github.com/robotology/gazebo-yarp-plugins/issues/61 and https://github.com/robotology/gazebo-yarp-plugins/issues/226 @traversaro


CMakeLists.txt
```diff
@@ -11,7 +11,11 @@ find_package(YARP 2.3.64.13 REQUIRED)
 # Finding dependencies
 find_package(Boost REQUIRED serialization system)
 find_package(Protobuf REQUIRED)
-find_package(Gazebo   REQUIRED)
+find_package(Gazebo REQUIRED)
+if (Gazebo_VERSION_MAJOR LESS 5.0)
```
traversaro: It should be sufficient to just have `find_package(Gazebo 5.0 REQUIRED)`, see https://cmake.org/cmake/help/v3.0/command/find_package.html .

traversaro: I said something stupid, as by default the version check is `SameMajorVersion` (https://bitbucket.org/osrf/gazebo/src/73cc987353ac777b6024dc1a4eac0b7602d4635e/CMakeLists.txt?at=default&fileviewer=file-view-default#CMakeLists.txt-423) that is clearly now what we want.

CMakeLists.txt
```diff
@@ -11,7 +11,11 @@ find_package(YARP 2.3.64.13 REQUIRED)
...
+    message(status "Gazebo version : " ${Gazebo_VERSION_MAJOR}.${Gazebo_VERSION_MINOR}.${Gazebo_VERSION_PATCH})
+    message(FATAL_ERROR "your gazebo version is older than GAZEBO 5.0. Gazebo Yarp plugins are supported with gazebo versions >= 5.0. Please update to a newer version")
```
traversaro: Capitalization needs to be fixed? `your` -> `Your` , `gazebo` -> `Gazebo`, 'GAZEBO` -> `Gazebo`.

Issue #221, opened 2015-12-16
Compilation is slow


traversaro:
Compilation of `gazebo-yarp-plugins` is extremely slow. 
This is probably due to the use of catch-them-all headers file of gazebo (such as [`<gazebo/physics/physics.hh>`](https://github.com/robotology/gazebo-yarp-plugins/search?utf8=%E2%9C%93&q=%3Cgazebo%2Fphysics%2Fphysics.hh%3E) or [`<gazebo/transport/transport.hh>`](https://github.com/robotology/gazebo-yarp-plugins/search?utf8=%E2%9C%93&q=%3Cgazebo%2Ftransport%2Ftransport.hh%3E&type=Code) ). 
Including only what we actually use could help to improve the compilation time.

traversaro:
A quick check with GCC's `-ftime-report` option seems still to indicate that the parser consume most of the time, while deprecation warnings does not seem to consume a big quantity of time.

traversaro:
Not a concrete issue for which we can say "solved" or not, as such I am closing it. If somebody wants to work fo improve the situation is anyway obviously welcome.

traversaro:
Related Gazebo issue : https://bitbucket.org/osrf/gazebo/issues/2426/simply-calling-gazebo-printversion .

PR #269, opened 2016-12-20
Catch them all headers cleanup

Yeshasvitvs:
See https://github.com/robotology/gazebo-yarp-plugins/issues/221
headers cleaned up from the all the plugins modules
@traversaro please check if this helps to reduce the compilation time

### DIFF
--- a/plugins/clock/src/Clock.cc
+++ b/plugins/clock/src/Clock.cc
@@ -9,7 +9,10 @@
 #include "ClockServerImpl.h"
 
 #include <gazebo/gazebo_config.h>
-#include <gazebo/physics/physics.hh>
+#include <gazebo/common/Events.hh>
+#include <gazebo/physics/PhysicsEngine.hh>
+#include <gazebo/physics/World.hh>
+#include <gazebo/physics/PhysicsIface.hh>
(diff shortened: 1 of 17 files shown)


traversaro:
This is great! I think this should reduce the compilation time, but even if this is not the case we should merge it anyway because it helps in cleaning the code. Thanks a lot @Yeshasvitvs !

Yeshasvitvs:
welcome @traversaro
if you also have someother ideas to reduce the compile time we can look into that

barbalberto:
:+1: 
What is taking a huge amount of compilation time is having all the warning check enabled.
We can reduce them to a reasonable check and enable all checks in the build farm or somewhere else.

francesco-romano:
I am not able to test this PR on macOS as I don't have gazebo installed anymore. Maybe @traversaro can do it?

@barbalberto i am completely against disabling warnings checks.

traversaro:
I think we can move the discussion on how to further speed-up the compilation process in https://github.com/robotology/gazebo-yarp-plugins/issues/221 . @Yeshasvitvs thanks for the PR, merging!

PR #280, opened 2017-02-21
lasersensor plugin requires yarp version 2.3.68

arocchi:
Fixes compilation for users with `YARP < 2.3.68`


traversaro:
Thanks, waiting for Travis to complete. If the ADVR projects are using a `YARP < 2.3.68` it would be better to use YARP 2.3.66 on some Travis configuration, to make sure to avoid breaking it again.

Issue #282, opened 2017-02-27
Gazebo 8  crashes if a model containing a `gazebo_yarp_controlboard` plugin is removed 


traversaro:
Gazebo crashes with no further information. 
Currently verified in Gazebo 8 on macOS Sierra by me and @nunoguedelha . 
Model removal seems to work fine on Gazebo 7 on Ubuntu 16.04 . 
We still need to test on Gazebo 8 on Ubuntu 16.04 .

traversaro:
Sometimes, on crash Gazeb outputs the following error:
~~~
gzserver(11133,0x700002e0d000) malloc: *** error for object 0x7febd0a73560: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
~~~

traversaro:
After half an hour of simulation:
~~~
[INFO]created device <gazebo_controlboard>. See C++ class GazeboYarpControlBoardDriver for documentation.
gzserver(11159,0x70000d6e5000) malloc: *** error for object 0x7f9d74d699f0: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
gzclient(11160,0x7fffa82af3c0) malloc: *** error for object 0x7f9dbcf64080: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
~~~

traversaro:
I think it is due to this change:
https://bitbucket.org/osrf/gazebo/src/cd526975bc8bdf3813e0600e3ece180870602471/Migration.md?at=default&fileviewer=file-view-default#Migration.md-631 .

traversaro:
More in detail, it seems that in Gazebo 8 the Disconnect methods are deprecated:
~~~
Deprecation: public: void Event::Disconnect(ConnectionPtr);
Deprecation: public: void EventT::Disconnect(ConnectionPtr);
Replacement: Delete the Connection object, perhaps by calling reset() on its smart pointer.
~~~
An easy fix for this can be to just called the `reset()` method on the connection after the call to the `Disconnect` method (this will continue to work for all Gazebo versions). The fact that the method is deprecated and not removed means that it should still work, and the fact that is not working I think is a Gazebo 8 bug. However I will not take the time to properly open a Gazebo bug because it is a transient behavior that only affect Gazebo 8 .

traversaro:
[...]

PR #285, opened 2017-02-28
Fix deletion of models using gazebo-yarp-plugins plugins in Gazebo 8

traversaro:
Streamlining of the disconnection event: simply resetting the
Connection shared pointer is enough to disconnect an event
since at least Gazebo 2. So the code was cleaned up from all
the call to the Disconnect* functions are all the unnecessary
check if a connection pointer is valid before resetting it.

Fix https://github.com/robotology/gazebo-yarp-plugins/issues/282 .

### DIFF
--- a/plugins/camera/src/CameraDriver.cpp
+++ b/plugins/camera/src/CameraDriver.cpp
@@ -148,8 +148,7 @@ bool GazeboYarpCameraDriver::close()
 {
     if (this->m_updateConnection.get())
     {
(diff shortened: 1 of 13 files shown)


traversaro:
To get a bit of a background on why the `Disconnect**` methods were actually never necessary, please check : https://bitbucket.org/osrf/gazebo_design/src/d5c73e6e0fef3711467335bde365c8a753cef14a/event_callbacks/event_callbacks.md?fileviewer=file-view-default , in particular the end of the "Lifecycle and Ownership" section.

PR #290, opened 2017-03-08
Add the gazebo_yarp_fakecontrolboard plugin

traversaro:
Plugin exposing a set of axes that are not actually simulated in Gazebo,
but that are present in the yarp controlboard always in controlmode not configured.

Useful for robots that need to expose a fixed set of joints in the controlboard,
even if those joints are not actually simulated.


traversaro:
Merging, additional comments can be done directly in the `devel` branch.

PR #295, opened 2017-04-07
Depth camera no subdevice

barbalberto:
Change depth Camera plugin in such a way that 'subdevice' keyword in '.ini' config file is not allowed.
Subdevice is always added in the singleton as a loaded plugin and reachable by all other plugins upon request.

CER config file has to be updated accordingly to keep working


barbalberto:
Removed shadowing pointer and rebased to latest devel
@traversaro

barbalberto:
@traversaro, is it ok to merge? First travis keep failing to set-up testing environment

traversaro:
I restarted the build, once it goes green I think it should be ok to merge!

barbalberto:
Ok, rebased over latest devel due to recent changes in yarp PID interface

traversaro:
Thanks!

Issue #303, opened 2017-06-12
Configuring incomplete, errors occurred!


emilio-cartoni:
After I updated to the latest commit, I get this error when I try to configure the build.
Last lines of `cmake ../ -DCMAKE_INSTALL_PREFIX=/home/emiliocartoni/opt/gazebo-yarp-plugins/build/`  look like this:
```
-- Looking for OGRE_Volume...
-- Looking for OGRE_Overlay...
-- Not processing thrift file clock_rpc.thrift, ALLOW_IDL_GENERATION=OFF
-- Not processing thrift file worldinterface_rpc.thrift, ALLOW_IDL_GENERATION=OFF
CMake Error at /usr/share/cmake-3.0/Modules/CMakePackageConfigHelpers.cmake:216 (message):
  Unknown keywords given to CONFIGURE_PACKAGE_CONFIG_FILE():
  "INSTALL_PREFIX;/home/emiliocartoni/opt/gazebo-yarp-plugins/build"
Call Stack (most recent call first):
  cmake/InstallBasicPackageFiles.cmake:306 (configure_package_config_file)
  CMakeLists.txt:71 (install_basic_package_files)

```

I tried to delete and recreate the build directory but it didn't help.

Apparently, the problem is located in the last lines of the CMakeLists.txt file (those added here: 94dc4c3a872e74a41b78dd6282f089d890cf3a80).

```
# Install the files necessary to call find_package(GazeboYarpPlugins) in CMake projects
include(InstallBasicPackageFiles)
install_basic_package_files(${PROJECT_NAME}
                            VARS_PREFIX ${PROJECT_NAME}
                            VERSION ${${PROJECT_NAME}_VERSION}
                            COMPATIBILITY AnyNewerVersion
                            TARGETS_PROPERTY ${PROJECT_NAME}_TARGETS
                            NO_CHECK_REQUIRED_COMPONENTS_MACRO)
```

If I delete them, everything works fine (cmake, make, make install).

traversaro:
Thanks for the issue, I think I know the source of this. 
Just to be sure, which version of CMake are you using?

emilio-cartoni:
cmake --version
```
cmake version 3.0.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```
Indeed, it's not 3.1 :)

traversaro:
Just out of curiosity, which OS are you using? Debian Jessie?

emilio-cartoni:
[...]

PR #306, opened 2017-06-15
Import CMakePackageConfigHelpers.cmake from iDynTree to fix configuration in CMake 3.0

traversaro:
Fix https://github.com/robotology/gazebo-yarp-plugins/issues/303 

@emilio-cartoni can you check this fixes your problem?


emilio-cartoni:
Yes, it did fix it, thanks :)

PR #314, opened 2017-06-30
[distributedinertials] Fix compilaton in Gazebo 9

traversaro:


traversaro:
What you mean by "Gazebo 9 is using the the format Vector3 again?" ? Gazebo 9 removed its internal math library in favor of https://bitbucket.org/ignitionrobotics/ign-math , and that's the reason that include needs to be removed, even because on previous releases of Gazebo it was not necessary.

aerydna:
merged thanks!

PR #321, opened 2017-10-03
Fix plugin worldinterface

aikolina:
In the previous version, it was not possible to get a link if its path was deeper than 2 levels. So it was possible to retrieve a link of the form `A::link`, but **NOT** one like `A::B::link`. This issue has been fixed here.


aerydna:
merged thanks!

PR #322, opened 2017-11-03
[controlboard] Add three-parameters variant of setRefTorques

traversaro:
This method was implemented in YARP (and in the ControlBoardWrapper)
in https://github.com/robotology/yarp/pull/349 , but it was never
implemented in Gazebo. Unfortunatly, as the setRefTorques result
in message sent only on the streaming port, the call to setRefTorques
was correctly returning true, and the only error was a print in the
ControlBoardWrapper running inside Gazebo, so the failure was quite silent.

I also cleanup the rest of methods of the ITorqueControl interface.

### DIFF
--- a/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
+++ b/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
@@ -229,18 +229,21 @@ public:
     bool changeInteractionMode(int j, yarp::dev::InteractionModeEnum mode); //private function
 
     //TORQUE CONTROL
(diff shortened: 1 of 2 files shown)


traversaro:
cc @diegoferigo @francesco-romano @gabrielenava @S-Dafarra this was not affecting all our controllers because the yarp-wholebodyinterface was not using the three-parameters setRefTorques . However the new WB-Toolbox 3.0 will use the `remotecontrolboardremapper`, and so it will be affected by this bug.

PR #324, opened 2017-11-14
Switch from using Gazebo internal JointController over Gazebo topics to embed a PID control in the gazebo_yarp_controlboard plugin

francesco-romano:
Fix https://github.com/robotology/gazebo-yarp-plugins/issues/165

### DIFF
--- a/plugins/controlboard/src/ControlBoardDriver.cpp
+++ b/plugins/controlboard/src/ControlBoardDriver.cpp
@@ -115,6 +115,7 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
     m_clock = 0;
     m_torqueOffset = 0;
 
+
     m_trajectory_generator.resize(m_numberOfJoints, NULL);
     m_coupling_handler.clear();
     m_speed_ramp_handler.resize(m_numberOfJoints, NULL);
@@ -338,15 +339,6 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
     this->m_resetConnection =
     gazebo::event::Events::ConnectWorldReset(boost::bind(&GazeboYarpControlBoardDriver::onReset, this));
 
-    m_gazeboNode = gazebo::transport::NodePtr(new gazebo::transport::Node);
-#if GAZEBO_MAJOR_VERSION >= 8
-    std::string worldName = this->m_robot->GetWorld()->Name(); 
-#else
-    std::string worldName = this->m_robot->GetWorld()->GetName(); 
-#endif
-    m_gazeboNode->Init(worldName);
-    m_jointCommandPublisher = m_gazeboNode->Advertise<gazebo::msgs::JointCmd>(std::string("~/") + this->m_robot->GetName() + "/joint_cmd");
-
     _T_controller = 1;
 
     resetPositionsAndTrajectoryGenerators();
@@ -453,8 +445,15 @@ bool GazeboYarpControlBoardDriver::configureJointType()
 
 void GazeboYarpControlBoardDriver::onUpdate(const gazebo::common::UpdateInfo& _info)
 {
+    //TODO: how to properly reset the initial time?
+    static bool firstTime = true;
+    if (firstTime) {
+        firstTime = false;
+        m_previousTime = _info.simTime;
+    }
+    gazebo::common::Time stepTime = _info.simTime - m_previousTime;
+    m_previousTime = _info.simTime;
     m_clock++;
-    //yDebug()<< deviceName << "1";
     // measurements acquisition
     for (size_t jnt_cnt = 0; jnt_cnt < m_jointPointers.size(); jnt_cnt++)
     {
@@ -468,7 +467,6 @@ void GazeboYarpControlBoardDriver::onUpdate(const gazebo::common::UpdateInfo& _i
         m_torques[jnt_cnt] = m_jointPointers[jnt_cnt]->GetForce(0u);
     }
 
-    //yDebug()<< deviceName << "2";
     //measurements decoupling
     for (size_t cpl_cnt = 0; cpl_cnt < m_coupling_handler.size(); cpl_cnt++)
     {
@@ -481,7 +479,6 @@ void GazeboYarpControlBoardDriver::onUpdate(const gazebo::common::UpdateInfo& _i
         }
     }
 
-    //yDebug()<< deviceName << "3";
     // check measured torque for hw fault
     for (size_t jnt_cnt = 0; jnt_cnt < m_jointPointers.size(); jnt_cnt++) {
         if (m_controlMode[jnt_cnt]!=VOCAB_CM_HW_FAULT && fabs(m_torques[jnt_cnt])>m_maxTorques[jnt_cnt])
@@ -493,8 +490,6 @@ void GazeboYarpControlBoardDriver::onUpdate(const gazebo::common::UpdateInfo& _i
 
     // Updating timestamp
     m_lastTimestamp.update(_info.simTime.Double());
-
-    //logger.log(m_velocities[2]);
 
     //update refernce m_jntReferenceVelocities
     for (size_t j = 0; j < m_numberOfJoints; ++j)
@@ -510,7 +505,6 @@ void GazeboYarpControlBoardDriver::onUpdate(const gazebo::common::UpdateInfo& _i
         }
     }
 
-    //yDebug()<< deviceName << "4";
     //update Trajectories
     for (size_t j = 0; j < m_numberOfJoints; ++j)
     {
@@ -542,7 +536,6 @@ void GazeboYarpControlBoardDriver::onUpdate(const gazebo::common::UpdateInfo& _i
         }
     }
 
-    //yDebug()<< deviceName << "5";
     //references decoupling
     m_motReferencePositions=m_jntReferencePositions;
     m_motReferenceVelocities=m_jntReferenceVelocities;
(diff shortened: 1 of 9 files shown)


francesco-romano:
This PR has potentially huge impact on all controllers and demo.
We should check and test in details the behaviour with this PR.
In particular:
- yoga (for torque control) ( @gabrielenava can you test it?)
- some position-based controllers
- some velocity-based controllers
- mixed/impedance?

traversaro:
@liesrock are you still using this plugins somewhere or are you only using XBotCore for simulation?

gabrielenava:
yes I can test it. Is this modification only related to simulations (so only concerns this repo)?

francesco-romano:
@gabrielenava yes. I changed how the plugins communicate with gazebo the references for position, velocity, torques, etc

gabrielenava:
I observed a strange behavior of iCub eyes. When I put the (old) iCub model inside Gazebo, this is what happens:

creepy_icub

gabrielenava:
Tested yoga, seesaw and standup demo in simulation. All of them were working as usual. I also did a comparison between the Yoga demo before and after this modification. No significative differences between the two demos were found. Data verified: torques, CoM, joint positions. 

To conclude, this PR passed the tests (but there is still the problem with the eyes).

traversaro:
@DanielePucci we need someone checking the eyes issue before merging.

DanielePucci:
Probably @prashanthr05 with the support of @nunoguedelha may investigate this issue

triccyx:
To be clear 
I'm currently use the decouple joint in the limit method and I think I should also use the limits decoulped as well. 
I mean: I want to translate the limits from ini file to the decouple version L R.

traversaro:
> I'm currently use the decouple joint in the limit method and I think I should also use the limits decoulped as well.
> I mean: I want to translate the limits from ini file to the decouple version L R.

I don't fully understand this point. Is the current version of the SDF model contained in icub-gazebo working with this PR without any modification to the .ini files?

triccyx:
At the moment the SDF contains the limit for Vs and Vg. But in the code the limits are checked with decouple joint. So this is not correct.

traversaro:
> At the moment the SDF contains the limit for Vs and Vg. But in the code the limits are checked with decouple joint. So this is not correct.

In the old version (before this PR) the limits were instead correctly checked on (Vs, Vg)? If this is the case, we should preserve this behaviour, and fix any kind of inconsistency in the parameter format in a separate PR.

triccyx:
At the end the file could be correct because it describes the limit of joint 4 and 5 that are Vs and Vg after the coupling line: 
[COUPLING]
eyes_vergence_control (4 5) (eyes_version eyes_vergence)

The problem is that I'm not sure if I can change and make work the limit function using the coupled joints and coupled limits, because the control is done using decoupled joint. And so the corrections due to limits problem should be done with decoupled joints position.
I don't know if I'm clear in this.
I try and tell you

traversaro:
I checked the code, and I think you are right, that code never worked as intended for coupled joints. The point is no one ever realized that because `check_joint_within_limits_override_torque` is only called for joint controlled in torque mode, and in our controllers (at least on iCub) we never control eyes and head in torque mode because on the real robot this is not supported.

triccyx:
Note that the if is commented
 //if (m_controlMode[i] == VOCAB_CM_TORQUE || m_interactionMode[i] == VOCAB_IM_COMPLIANT)

traversaro:
I understood what changed. `check_joint_within_limits_override_torque` used to be called only for when the control mode was TORQUE and the interaction mode was COMPLIANT, because it was called inside the `sendTorqueToGazebo` method`, but now it is instead called for all control modes because it is called in https://github.com/robotology/gazebo-yarp-plugins/pull/324/files#diff-a01901eeedeed61dae3fabb62c01e331R608 . This exposed the problem about coupled joints.

triccyx:
So can I uncomment the if?

traversaro:
I think that would be a clean solution (perhaps adding the `m_controlMode[i] == VOCAB_CM_PWM` and `m_controlMode[i] == VOCAB_CM_CURRENT` condition to match the old behavior).

triccyx:
I'm aligned with Marco R.

triccyx:
Relating to:
std::vector<Range> m_hardwarePosLimits; //**< hardware limits from relal motors*/
I check the code and the suggestion to use m_jointPointers[i]->UpperLimit
doesn't work as a matter of fact already
m_jointPosLimits[i].max = convertGazeboToUser(i, m_jointPointers[i]->UpperLimit(0));

For this PR I wouldn't include this m_hardwarePosLimits.
For a new PR:
1)check limits for positions control
2)m_hardwarePosLimits
3)check formula for VG and VS calculation

Are you agree?

triccyx:
Also I don't like that the var m_position is filled as motor position and then became joint position

traversaro:
> Also I don't like that the var m_position is filled as motor position and then became joint position 

I totally agree on this, but I think this modification is better suite for a new PR.

plugins/controlboard/.vscode/settings.json
```diff
@@ -0,0 +1,40 @@
...
+        "ratio": "cpp",
+        "stdexcept": "cpp",
+        "system_error": "cpp",
+        "thread": "cpp",
+        "tuple": "cpp",
+        "type_traits": "cpp",
+        "typeinfo": "cpp",
+        "utility": "cpp"
+    }
+}
```
traversaro: @triccyx Can you remove this unrelated addition?

plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -1270,36 +1270,38 @@ bool GazeboYarpControlBoardDriver::check_joint_within_limits_override_torque(int
 
     int   signKp = (positionPID.GetPGain()>0?1:-1);
     int   signRef = (ref>0?1:-1);
-    //if (m_controlMode[i] == VOCAB_CM_TORQUE || m_interactionMode[i] == VOCAB_IM_COMPLIANT)
-    if (m_controlMode[i] != VOCAB_CM_IDLE && m_controlMode[i] != VOCAB_CM_FORCE_IDLE)
+    if (m_controlMode[i] == VOCAB_CM_TORQUE || m_interactionMode[i] == VOCAB_IM_COMPLIANT)
```
traversaro: To be consistent with the code before this PR, the check should be active also for `VOCAB_CM_PWM` and `VOCAB_CM_CURRENT`, thanks!

plugins/controlboard/.vscode/settings.json
triccyx: Ok

plugins/controlboard/src/ControlBoardDriver.cpp
triccyx: Do you mean :
if ((m_controlMode[i] == VOCAB_CM_TORQUE || m_interactionMode[i] == VOCAB_IM_COMPLIANT)  ||  (m_controlMode[i] == VOCAB_CM_PWM || m_interactionMode[i] == VOCAB_CM_CURRENT)

traversaro: Yes, given that the logical or is an operation for which the associative property hold, we can just put only a couple of parenthesis.

traversaro:
Once you merge the last two commits the PR is good to go for me. 
@S-Dafarra @gabrielenava may want to briefly try this branch before we merge it. 
I suggest that we also wait for @randaz81 approval for merging.

S-Dafarra:
Seems fine to me, thanks!

traversaro:
> They should be there

I still see the two separate commits https://github.com/robotology/gazebo-yarp-plugins/pull/324/commits/05ec8444e26a972de78e6364e9f84d35aef1828e and https://github.com/robotology/gazebo-yarp-plugins/pull/324/commits/c99c476c9af87cea904c9ea7908498aa38c3b915 but it is not a big problem, approving the PR.

gabrielenava:
tested, fine for me!

traversaro:
@triccyx  I have asked @randaz81 and he told me he will review this PR in the next couples of days. If he does not, we are authorized to ping him on this review. : )

traversaro:
cc @triccyx some conflicts emerged after merging https://github.com/robotology/gazebo-yarp-plugins/pull/334 . Can you rebase the branch w.r.t. to devel and fix the conflicts? Thanks!

traversaro:
@triccyx It seems that the branch is not compiling anymore with the following error (see https://travis-ci.org/robotology/gazebo-yarp-plugins/jobs/327305894):
~~~
/app/plugins/controlboard/src/ControlBoardDriver.cpp:24:64: error: initializer 'deviceName' does not name a non-static data member or base class; did you mean the member 'm_deviceName'?
GazeboYarpControlBoardDriver::GazeboYarpControlBoardDriver() : deviceName("") {}
                                                               ^~~~~~~~~~
                                                               m_deviceName
/app/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h:364:17: note: 'm_deviceName' declared here
    std::string m_deviceName;
~~~

triccyx:
I have done the fix then commit and push. But perhaps do you prefer that I redo the rebase procedure from beginning?

traversaro:
Yes, for being able to use [`git bisect`](https://git-scm.com/docs/git-bisect) in the future, it is important to keep the history free of "broken" commits that do not compile, and hence it is typically a good idea to rebase to ensure that all the commits merged are good (i.e. they don't have compile errors).

triccyx:
If I try:
git rebase origin/devel
on my local branch
<Current branch changed_pids is up to date.>
What am I doing wrong?

traversaro:
@triccyx  Can you squash the last 4 or 5 commits? Thanks!

traversaro:
I was able to rebase ignoring the fact that the last three commits where not displayed in the rebase list, and everything went smoothly. I still don't know what was going on, but I think it was due to some strange duplicate commit going on. For historical reference, the commit is this one: https://github.com/robotology/gazebo-yarp-plugins/commit/1b6acc440ef7ffb1375d43c077a9b35ddf1e7a9e .

Issue #325, opened 2017-12-09
Inserting in Gazebo the new model iCubGazeboV2_5 directly from icub-models crashes the simulator


nunoguedelha:
This issue only occurs if we insert in Gazebo the new model iCubGazeboV2_5 directly from the `icub-models` repo. This is probably due to an issue with the nesting of gazebo models and the resulting path to inertial sensors.
The current implementation of the plugins handling the inertial MTB sensors (`libgazebo_yarp_inertial.so` and `libgazebo_yarp_distributedinertials.so`) were designed considering a model nested like in the case of `icub_fixed_no_hands`, and might have a faulty behavior when the robot model is not nested.

nunoguedelha:
Issue moved from [robotology-playground/icub-models #10](https://github.com/robotology-playground/icub-models/issues/10).

nunoguedelha:
### Crash analysis
As soon as the robot is dropped in the simulator environment, the simulator crashes throwing the error:
```
[DEBUG]Initializing Trajectory Generator with current values 
[INFO]created device <gazebo_controlboard>. See C++ class GazeboYarpControlBoardDriver for documentation.
yarp: Port /icubSim/applyExternalWrench/rpc:i active at tcp://192.168.1.101:10049/
gzclient(58718,0x7fffcaf0f3c0) malloc: *** error for object 0x7fd9dc05d2b0: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
```
If we look at the first whole log messages, we can see:
```
yarp: Port /icubSim/left_leg/inertialMTB/rpc:i active at tcp://192.168.1.101:10015/
[INFO]AnalogServer : no ROS initialization required 
[INFO]created wrapper <analogServer>. See C++ class yarp::dev::AnalogWrapper for documentation.
[ERROR]GazeboYarpInertialMTBPartDriver Error: required sensor iCub::l_upper_leg::l_upper_leg_mtb_acc_10b1  was not found 
[INFO]created device <gazebo_inertialMTB>. See C++ class GazeboYarpInertialMTBPartDriver for documentation.
yarp: Port /icubSim/left_leg/inertialMTB active at tcp://192.168.1.101:10016/
[INFO]No ROS group found in config file ... skipping ROS initialization. 
yarp: Port /icubSim/right_leg/inertialMTB/rpc:i active at tcp://192.168.1.101:10017/
...
```
[...]

PR #326, opened 2017-12-09
Fixes issue 325 (mismatch between inertial sensors scoped names from the ini config files and sensor plugins)

nunoguedelha:
Fix the mismatch between the inertial sensors scoped names set from the ini config files and those registered by the sensor plugins. 
Fixes issue https://github.com/robotology/gazebo-yarp-plugins/issues/325


traversaro:
`devel` branch added, even if I fail to understand why "Not Analyzed == Failed".

nunoguedelha:
As soon as I click "reanalyze", I get an error ("error occured"), and I weren't able to see any log.

PR #327, opened 2017-12-15
Added check on torque validity

triccyx:


traversaro:
Thanks, commenting using reviewer feature of github.

plugins/controlboard/src/ControlBoardDriverTorqueControl.cpp
```diff
@@ -6,12 +6,17 @@
...
 #include "ControlBoardDriver.h"
+#include <yarp/os/Log.h>
+#include <yarp/os/LogStream.h>
 
 
 using namespace yarp::dev;
 
 bool GazeboYarpControlBoardDriver::setRefTorque(int j, double t)
 {
+    if(!checkIfTorqueIsValid(t))
```
traversaro: Nitpick: while we are still finalize the style used in this repositories, we typically keep a space between operators such as `if` and `for` and the parenthesis in C++ code, to differentiate them from function calls, in which there is not space instead.

traversaro:
Thanks, merging the PR.

PR #329, opened 2017-12-20
Fix eye

triccyx:
Fix couple joints problems and minor name refactoring.
Be aware to check .ini file for eyes L, R limits


plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -549,7 +552,7 @@ void GazeboYarpControlBoardDriver::onUpdate(const gazebo::common::UpdateInfo& _i
         if ((m_controlMode[j] == VOCAB_CM_POSITION || m_controlMode[j] == VOCAB_CM_POSITION_DIRECT) && (m_interactionMode[j] == VOCAB_IM_STIFF))
         {
             gazebo::common::PID &pid = m_pids[VOCAB_PIDTYPE_POSITION][j];
-            forceReference = pid.Update(convertUserToGazebo(j, m_positions[j]) - convertUserToGazebo(j, m_motReferencePositions[j]), stepTime);
+            forceReference = pid.Update(convertUserToGazebo(j, m_positionsDecoupled[j]) - convertUserToGazebo(j, m_motReferencePositions[j]), stepTime);
```
traversaro: Please avoid adding trailing whitespace in the code, that complicates reviewing your modification (see http://www.dinduks.com/why-are-trailing-whitespaces-bad/). Thanks!

traversaro:
Hi @triccyx , can you update the PR https://github.com/robotology/gazebo-yarp-plugins/pull/324 instead of opening a new one? It should be sufficient to just push your commits to the `changed_pids` branch.

 Regarding the problem in the configuration files of the eyes, can you please open an issue in the icub-gazebo repository explaining the problem? Or even better it would be an update of the configuration files via a Pull Request to the icub-gazebo repository.

traversaro:
@triccyx minor: your commit seems to be still using the email luca.tricerri@iit.it , that GitHub does not recognize.

plugins/controlboard/src/ControlBoardDriver.cpp
triccyx: ok!

triccyx:
About limits can you suggest me L and R correct max and min ? Or foreward me to the right person to ask?

traversaro:
> About limits can you suggest me L and R correct max and min ? Or foreward me to the right person to ask?

On a second thought, even if the semantics of the limits is not really clear, the behaviour of the plugin should not change. Perhaps the `check_joint_within_limits_override_torque` function should be modified to use `m_positionDecoupled` as well?

traversaro:
Minor advice (not a big problem for this PR, but for the future): for simplify the review, it is typically a good idea to separate commits that modify the logic of the source code from commits that only do renaming and other stylistics changes.

triccyx:
Yes I was already modify the limits method in that way! I think we are in the same wave, ahahaha!
About commits ok, I was just trying the rebase to obtain one commit instead of two but this is not the correct case, sorry.

traversaro:
Great, let's continue the discussion on the other PR.

Issue #87, opened 2014-04-15
Migrate all the printf/cout to gazebo defined macros


francesco-romano:
Instead of using `printf` or `cout` we should use gazebo defined output functions

Gazebo defines the following loggers:
- `gzdbg`
- `gzerr`
- `gzwarn`

They are defined in `common`: `gazebo/common/Console.hh`

traversaro:
Even for yarp output that is not directly sent to gazebo streams (such as information on port opened, file searched by resource finder) we should enable/disable the verbosity using the [`Console::GetQuiet()`](https://osrf-distributions.s3.amazonaws.com/gazebo/api/dev/classgazebo_1_1common_1_1Console.html#a1240d0b06e31d05ee891e3880224c06f) method to get if Gazebo was launched with the `--verbose` option (typically used to understand when something is not working).

traversaro:
Most of the prints moved to YARP logger in https://github.com/robotology/gazebo-yarp-plugins/pull/242 .

PR #334, opened 2018-01-08
Removing hardcoded parameters in gazebo_yarp_controlboard plugin 

### LINKED ISSUE #332: Remove hardcoded parameters in controlboard plugin
There are a few hardcoded parameters in the controlboard plugin, including:
* The `maxTorques` vector, used to decide when the joint goes in hardware fault mode, hardcoded to 2000 ( https://github.com/robotology/gazebo-yarp-plugins/blob/master/plugins/controlboard/src/ControlBoardDriver.cpp#L69 , see https://github.com/robotology/gazebo-yarp-plugins/issues/331  )

* There are also the parameters:
~~~
const double RobotPositionTolerance_revolute = 0.9;      // Degrees
const double RobotPositionTolerance_linear = 0.004; // Meters
~~~
but with a preliminary check it seems to be that those values are unused. In that case they can be directly removed.

triccyx:
Fix #332

### DIFF
--- a/plugins/controlboard/src/ControlBoardDriver.cpp
+++ b/plugins/controlboard/src/ControlBoardDriver.cpp
@@ -21,9 +21,6 @@ using namespace yarp::sig;
 using namespace yarp::dev;
 
 
-const double RobotPositionTolerance_revolute = 0.9;      // Degrees
-const double RobotPositionTolerance_linear   = 0.004;    // Meters
-
 GazeboYarpControlBoardDriver::GazeboYarpControlBoardDriver() : deviceName("") {}
 
 GazeboYarpControlBoardDriver::~GazeboYarpControlBoardDriver() {}
@@ -302,6 +299,24 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
         if (m_trajectoryGenerationReferenceSpeed[j] >  m_jointVelLimits[j].max) m_trajectoryGenerationReferenceSpeed[j]= m_jointVelLimits[j].max/10;
     }
 
+    if (!setPositionsToleranceLinear())
+    {
+        yError()<<"Failed PositionsToleranceLinear initialization";
+        return false;
+    }
+
+    if (!setPositionsToleranceRevolute())
+    {
+        yError()<<"Failed PositionsToleranceRevolute initialization";
+        return false;
+    }
+
+    if (!setMaxTorques())
+    {
+        yError()<<"Failed Max Torque initialization";
+        return false;
+    }
+
     if (!setMinMaxImpedance())
     {
         yError()<<"Failed Impedance initialization";
@@ -419,14 +434,14 @@ bool GazeboYarpControlBoardDriver::configureJointType()
             case ( gazebo::physics::Entity::HINGE_JOINT  |  gazebo::physics::Entity::JOINT):
             {
                 m_jointTypes[i] = JointType_Revolute;
-                m_positionThreshold[i] = RobotPositionTolerance_revolute;
+                m_positionThreshold[i] = m_robotPositionToleranceRevolute;
                 break;
             }
 
             case ( gazebo::physics::Entity::SLIDER_JOINT |  gazebo::physics::Entity::JOINT):
             {
                 m_jointTypes[i] = JointType_Prismatic;
-                m_positionThreshold[i] = RobotPositionTolerance_linear;
+                m_positionThreshold[i] = m_robotPositionToleranceLinear;
                 break;
             }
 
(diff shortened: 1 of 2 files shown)


plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -1144,13 +1159,120 @@ bool GazeboYarpControlBoardDriver::setPIDsForGroup_IMPEDANCE(std::vector<std::st
     }
 }
 
-bool GazeboYarpControlBoardDriver::setMinMaxImpedance()
+bool GazeboYarpControlBoardDriver::FindMotorControlGroup(yarp::os::Bottle& motorControlGroup_bot) const
```
traversaro: For consistency I would keep method names initials lowercase.

plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -1144,13 +1159,120 @@ bool GazeboYarpControlBoardDriver::setPIDsForGroup_IMPEDANCE(std::vector<std::st
...
+    if (!FindMotorControlGroup(kin_chain_bot))
+        return false;
+
+    if (!kin_chain_bot.check("positionToleranceLinear")) {
+        yWarning()<<"No positionToleranceLinear value found in ini file, default one will be used!";
+        yDebug()<<"positionToleranceLinear: [ "<<m_robotPositionToleranceLinear<<" ]";
+        return true;
+    }
+
+    yInfo()<<"positionToleranceLinear param found!";
```
traversaro: I think we can remove this print.

plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -1144,13 +1159,120 @@ bool GazeboYarpControlBoardDriver::setPIDsForGroup_IMPEDANCE(std::vector<std::st
...
+
+bool GazeboYarpControlBoardDriver::setPositionsToleranceLinear()
+{
+    yarp::os::Bottle kin_chain_bot;
+    if (!FindMotorControlGroup(kin_chain_bot))
+        return false;
+
+    if (!kin_chain_bot.check("positionToleranceLinear")) {
+        yWarning()<<"No positionToleranceLinear value found in ini file, default one will be used!";
+        yDebug()<<"positionToleranceLinear: [ "<<m_robotPositionToleranceLinear<<" ]";
```
traversaro: If we intend to leave this in the code, I would turn it in a warning for consistency with the previous message, otherwise I would remove it.

plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -1144,13 +1159,120 @@ bool GazeboYarpControlBoardDriver::setPIDsForGroup_IMPEDANCE(std::vector<std::st
...
+    }
+
+    yInfo()<<"positionToleranceLinear param found!";
+    yarp::os::Bottle& positionToleranceLinear_bot = kin_chain_bot.findGroup("positionToleranceLinear");
+    if (static_cast<size_t>(positionToleranceLinear_bot.size()) != 2) {
+        yError()<<"Invalid number of params:positionToleranceLinear:"<<static_cast<size_t>(positionToleranceLinear_bot.size());
+        return false;
+    }
+
+    auto tmp=positionToleranceLinear_bot.get(1);
```
traversaro: I think here we can replace auto with `yarp::os::Value` for the sake of readability.

plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -1144,13 +1159,120 @@ bool GazeboYarpControlBoardDriver::setPIDsForGroup_IMPEDANCE(std::vector<std::st
...
+    if (!tmp.isDouble())
+    {
+        yError()<<"Invalid param type:positionToleranceLinear";
+        return false;
+    }
+
+    m_robotPositionToleranceLinear=tmp.asDouble();       
+
+    yDebug()<<"positionToleranceLinear: [ "<<m_robotPositionToleranceLinear<<" ]";
```
traversaro: I think we can remove this print.

plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -1144,13 +1159,120 @@ bool GazeboYarpControlBoardDriver::setPIDsForGroup_IMPEDANCE(std::vector<std::st
...
+    if (!FindMotorControlGroup(kin_chain_bot))
+        return false;
+
+    if (!kin_chain_bot.check("positionToleranceRevolute")) {
+        yWarning()<<"No positionToleranceRevolute value found in ini file, default one will be used!";
+        yDebug()<<"positionToleranceRevolute: [ "<<m_robotPositionToleranceRevolute<<" ]";
+        return true;
+    }
+
+    yInfo()<<"positionToleranceRevolute param found!";
```
traversaro: I think we can remove this print.

plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -1144,13 +1159,120 @@ bool GazeboYarpControlBoardDriver::setPIDsForGroup_IMPEDANCE(std::vector<std::st
...
+    auto tmp=positionToleranceRevolute_bot.get(1);
+    if (!tmp.isDouble())
+    {
+        yError()<<"Invalid param type:positionToleranceRevolute";
+        return false;
+    }
+
+    m_robotPositionToleranceRevolute=tmp.asDouble();       
+
+    yDebug()<<"positionToleranceRevolute: [ "<<m_robotPositionToleranceRevolute<<" ]";
```
traversaro: I think we can remove this print.

plugins/controlboard/src/ControlBoardDriver.cpp
triccyx: Ok

plugins/controlboard/src/ControlBoardDriver.cpp
triccyx: I remove

plugins/controlboard/src/ControlBoardDriver.cpp
triccyx: Ok

plugins/controlboard/src/ControlBoardDriver.cpp
triccyx: Ok

plugins/controlboard/src/ControlBoardDriver.cpp
triccyx: Ok

plugins/controlboard/src/ControlBoardDriver.cpp
triccyx: I prefer to keep, also because is present also in the other methods

plugins/controlboard/src/ControlBoardDriver.cpp
triccyx: I prefer to keep, also because is present also in the other methods

plugins/controlboard/src/ControlBoardDriver.cpp
traversaro: I see, I would want to remove also the others because the output of Gazebo is extremly confusing but you are right, consistency wins. : ) 
We should eventually move to have some verbosity option to enable/disable this prints at runtime, either using the `verbose` option of the YARP device or directly the Gazebo verbosity level. 
See https://github.com/robotology/gazebo-yarp-plugins/issues/87 and https://github.com/robotology/gazebo-yarp-plugins/pull/263 .

plugins/controlboard/src/ControlBoardDriver.cpp
traversaro: I see, ref https://github.com/robotology/gazebo-yarp-plugins/pull/334#discussion_r160383752 .

traversaro:
Merging as soon as Travis turns green.

S-Dafarra:
Sorry for bothering, can you provide an explanation of what is going to be merged here :slightly_smiling_face:. At least for posterity.

S-Dafarra:
Is it for solving issue https://github.com/robotology/gazebo-yarp-plugins/issues/332 ? (I saw it too late :sweat_smile: )

traversaro:
@triccyx what @S-Dafarra points out make sense, it is probably a good idea to use a more descriptive PR name, something like "Removing hardcoded parameters in gazebo_yarp_controlboard plugin" is probably ok!

triccyx:
Ok but keep also #332 somewhere

traversaro:
""Removing hardcoded parameters in gazebo_yarp_controlboard plugin #332"

S-Dafarra:
> Otherwise a link to the bug is missing

You can also put it in the description of the PR :wink:

PR #338, opened 2018-01-12
Sensorplugin/contactpressuresensorsarray

prashanthr05:
This device is used to simulate the contact normal forces that might be observed by an arbitrary number of load cells at specific locations with respect to a given link's origin frame. 
It uses the instance of a Gazebo Contact Sensor which provides contact information of the link (specifically providing wrenches acting at the CG). This information is transformed and mapped to the desired contact normal forces. In its current version, this device reads the location of the load cells (to be mentioned manually from the urdf or sdf) from a configuration file. 

- Reads the contact forces and torques acting at the gazebo collision level through a  gazebo ContactSensor instance. 
- Finds an equivalent wrench acting at the CG of the specified link.
- Transforms this equivalent wrench to the origin of the specified link. This transformation is necessary since the load cell locations are usually designed as fixed joints  with respect to the link origin in the model. 
- Finally, the transformed equivalent wrench is mapped to an array of contact normal forces acting at the load cell locations.  This mapping problem is an infinite solution problem and is solved by obtaining a pseudo inverse,  which gives the results with a minimized norm.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -17,3 +17,4 @@ Marco Randazzo <marco.randazzo@iit.it>
 Miguel Aragao <miguelaragao91@gmail.com>
 Mirko Ferrati <mirko.ferrati@gmail.com>
 Silvio Traversaro <silvio.traversaro@iit.it>
+Prashanth Ramadoss <prashanth.ramadoss@iit.it>
--- a/doc/embed_plugins.md
+++ b/doc/embed_plugins.md
@@ -30,7 +30,7 @@ but are nevertheless useful for simulating complex scenarios.
 | Lidar Sensor        | `gazebo_yarp_lasersensor` | Sensor | gazebo::GazeboYarpLaserSensor | yarp::dev::GazeboYarpLaserSensorDriver |
 | Coupled Joint Encoders | `gazebo_yarp_maissensor` | Model |  gazebo::GazeboYarpMaisSensor | yarp::dev::GazeboYarpMaisSensorDriver |
 | Stereo Cameras         | `gazebo_yarp_multicamera` | Sensor | gazebo::GazeboYarpMultiCamera | yarp::dev::GazeboYarpMultiCameraDriver |
-| Contact Load Cell Array| `gazebo_contactloadcellarray` | Model (check documentation for this choice)| gazebo::GazeboYarpContactLoadCellArray| yarp::dev::GazeboYarpContactLoadCellArrayDriver|
+| Contact Load Cell Array| `gazebo_contactloadcellarray` | Model | gazebo::GazeboYarpContactLoadCellArray | yarp::dev::GazeboYarpContactLoadCellArrayDriver |
 
 ## Plugins exposing simulation-specific functionalities
 |  Functionality     | Plugin Name  | Plugin Type |  Gazebo Plugin class  |
--- a/doc/embed_plugins.md
+++ b/doc/embed_plugins.md
@@ -12,7 +12,7 @@ For additional information it is possible to access the official documentation f
 
 The `gazebo-yarp-plugins` consists of:
 * gazebo plugins that instantiate yarp device drivers and
-* yarp device drivers that wrap gazebo functionalities inside the yarp device interfaces . 
+* yarp device drivers that wrap gazebo functionalities inside the yarp device interfaces .
 
 Furthermore, in `gazebo-yarp-plugins` there are several plugins that do not mimic any interface or capability present in real robots,
 but are nevertheless useful for simulating complex scenarios.
@@ -30,6 +30,7 @@ but are nevertheless useful for simulating complex scenarios.
 | Lidar Sensor        | `gazebo_yarp_lasersensor` | Sensor | gazebo::GazeboYarpLaserSensor | yarp::dev::GazeboYarpLaserSensorDriver |
 | Coupled Joint Encoders | `gazebo_yarp_maissensor` | Model |  gazebo::GazeboYarpMaisSensor | yarp::dev::GazeboYarpMaisSensorDriver |
 | Stereo Cameras         | `gazebo_yarp_multicamera` | Sensor | gazebo::GazeboYarpMultiCamera | yarp::dev::GazeboYarpMultiCameraDriver |
+| Contact Load Cell Array| `gazebo_contactloadcellarray` | Model (check documentation for this choice)| gazebo::GazeboYarpContactLoadCellArray| yarp::dev::GazeboYarpContactLoadCellArrayDriver|
 
 ## Plugins exposing simulation-specific functionalities
 |  Functionality     | Plugin Name  | Plugin Type |  Gazebo Plugin class  |
--- a/plugins/CMakeLists.txt
+++ b/plugins/CMakeLists.txt
@@ -44,3 +44,4 @@ else()
     message(STATUS "Gazebo version is lower than 7 or YARP version is lower than 2.3.67, disabling depthCamera  plugin.")
 endif()
 
+add_subdirectory(contactloadcellarray)
--- a/plugins/contactloadcellarray/CMakeLists.txt
+++ b/plugins/contactloadcellarray/CMakeLists.txt
@@ -1,0 +1,23 @@
+# Copyright (C) 2018 Istituto Italiano di Tecnologia ADVR & iCub Facility & RBCS Department
+# Authors: see AUTHORS file.
+# CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
+
+cmake_minimum_required(VERSION 2.8.7)
+
+include(AddGazeboYarpPluginTarget)
+
+set(LIB_COMMON_NAME gazebo_yarp_lib_common)
+if(CMAKE_VERSION VERSION_LESS 3.0.0)
+  get_property(GAZEBO_YARP_COMMON_HEADERS GLOBAL PROPERTY GAZEBO_YARP_COMMON_HEADERS)
+  unset(LIB_COMMON_NAME)
+endif()
+
+add_gazebo_yarp_plugin_target(LIBRARY_NAME contactloadcellarray
+			      INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR}
+			      INCLUDE_DIRS include/gazebo include/yarp/dev
+			      SYSTEM_INCLUDE_DIRS ${GAZEBO_YARP_COMMON_HEADERS} ${YARP_INCLUDE_DIRS} ${Boost_INCLUDE_DIRS} ${GAZEBO_INCLUDE_DIRS} ${SDFORMAT_INCLUDE_DIRS} ${PROTOBUF_INCLUDE_DIRS}
+			      LINKED_LIBRARIES ${LIB_COMMON_NAME} gazebo_yarp_singleton ${YARP_LIBRARIES} ${GAZEBO_LIBRARIES} ${Boost_LIBRARIES} ${SDFORMAT_LIBRARIES}
+			      HEADERS include/gazebo/ContactLoadCellArray.hh
+				      include/yarp/dev/ContactLoadCellArrayDriver.h
+			      SOURCES src/ContactLoadCellArray.cc
+				      src/ContactLoadCellArrayDriver.cpp)
--- a/plugins/contactloadcellarray/include/gazebo/ContactLoadCellArray.hh
+++ b/plugins/contactloadcellarray/include/gazebo/ContactLoadCellArray.hh
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2018 Fondazione Istituto Italiano di Tecnologia RBCS & iCub Facility & ADVR
+ * Authors: see AUTHORS file.
+ * CopyPolicy: Released under the terms of the LGPLv2.1 or any later version, see LGPL.TXT or LGPL3.TXT
+ */
+
+#ifndef GAZEBOYARP_CONTACTLOADCELLARRAY_HH
+#define GAZEBOYARP_CONTACTLOADCELLARRAY_HH
+
+#include <gazebo/common/Plugin.hh>
+#include <string>
+#include <yarp/dev/PolyDriverList.h>
+
+#include <sdf/sdf_config.h>
+#include <sdf/Element.hh>
+#include <sdf/Param.hh>
+
+
+namespace yarp 
+{
+    namespace dev
+    {
+        class IMultipleWrapper;
+    }
+}
+
+namespace gazebo
+{
+    class GazeboYarpContactLoadCellArray : public ModelPlugin
+    {
+        public:
+        GazeboYarpContactLoadCellArray();
+        virtual ~GazeboYarpContactLoadCellArray();
+    
+       /**
+        * Saves the gazebo pointer, creates the device driver
+        */
+        void Load(physics::ModelPtr _parent, sdf::ElementPtr _sdf);
+    
+        private:
+       
+       /**
+        * Polydriver instance to connect to analog sensor server
+        */
+        yarp::dev::PolyDriver m_devWrapper;
+        
+       /**
+        * Wrapper to attach all necessary devices
+        */
+        yarp::dev::IMultipleWrapper *m_imultwrapper;
+        
+       /**
+        * Polydriver instance to instantiate the device driver
+        */
+        yarp::dev::PolyDriver m_devDriver;
+    
+       /**
+        * YARP property to store information from config file
+        */
+        yarp::os::Property m_parameters;    
+        
+       /**
+        * Network interface to connect to
+        */
+        std::string m_wrapperName;
+        
+       /**
+        * Device driver name to be instantiated
+        */
+        std::string m_sensorDriverName;
+ 
+       /**
+        * Name of the robot model loaded by model plugin 
+        */
+        std::string m_gazeboRobotName;
+   
+    };
+}
+
+
+#endif 
--- a/plugins/contactloadcellarray/include/yarp/dev/ContactLoadCellArrayDriver.h
+++ b/plugins/contactloadcellarray/include/yarp/dev/ContactLoadCellArrayDriver.h
@@ -1,0 +1,230 @@
+/*
+ * Copyright (C) 2018 Fondazione Istituto Italiano di Tecnologia RBCS & iCub Facility & ADVR
+ * Authors: see AUTHORS file.
+ * CopyPolicy: Released under the terms of the LGPLv2.1 or any later version, see LGPL.TXT or LGPL3.TXT
+ */
+
+#ifndef GAZEBOYARP_CONTACTLOADCELLARRAYDRIVER_H
+#define GAZEBOYARP_CONTACTLOADCELLARRAYDRIVER_H
+
+#include <yarp/dev/DeviceDriver.h>
+#include <yarp/dev/IAnalogSensor.h>
+#include <yarp/dev/PreciselyTimed.h>
+#include <yarp/os/Stamp.h>
+#include <yarp/os/Semaphore.h>
+
+#include <yarp/sig/Matrix.h>
+
+#include <gazebo/common/Plugin.hh>
+#include <gazebo/common/common.hh>
+
+
+#include <gazebo/sensors/Sensor.hh>
+#include <gazebo/sensors/ContactSensor.hh>
+#include <gazebo/sensors/SensorManager.hh>
+
+#include <gazebo/physics/Contact.hh>
+#include <gazebo/physics/Link.hh>
+#include <gazebo/physics/Inertial.hh>
+#include <gazebo/physics/physics.hh>
+
+#include <ignition/math/Pose3.hh>
+#include <ignition/math/Vector3.hh>
+#include <ignition/math/Matrix3.hh>
+#include <ignition/math.hh>
+
+namespace yarp
+{
+    namespace dev
+    {
+        class GazeboYarpContactLoadCellArrayDriver;
+    }
+}
+
+/**
+ * YARP Device Driver for an array of contact load cells giving 1-axis force at different locations
+ * 
+ * This device is used to simulate the contact normal forces that might be 
+ * measured by an arbitrary number of load cells at specific locations with respect
+ * to a given link's origin frame. It uses the instance of a Gazebo Contact Sensor which
+ * provides contact information of the link (specifically providing wrenches - 
+ * 6 axis force torque components,  acting at different position with respect to the link).
+ * The equivalent wrench acts on the Center of Gravity(CG) of the link expressed in link origin frame.
+ * The CG of thee link is defined by the <inertial > in the sdf as a child of <link> tag.
+ * This information is transformed and mapped to the desired contact normal forces.
+ * 
+ * MAJOR ASSUMPTION: The load cells are distributed over a plane perpendicular to the z-axis of the link frame
+ * and the load cells are measuring the force over the z-direction
+ * 
+ * In its current version, this device reads the location of the load cells (to be copied manually from the urdf or sdf onto the configuration file)
+ * from a configuration file. Reads the contact forces and torques acting at the gazebo collision level through a 
+ * gazebo ContactSensor instance. Finds an equivalent wrench acting at the CG of the specified link(location CG, rotation same as link origin). Transforms this
+ * equivalent wrench to the origin of the specified link. This transformation is necessary since the load cell locations are
+ * with respect to the link origin. Finally, the transformed equivalent wrench is mapped to an array of contact normal forces acting
+ * at the load cell locations. 
+ * 
+ * This mapping problem is an infinite solution problem and is solved by obtaining a pseudo inverse, 
+ * which gives the results with a minimized norm.
+ * 
+ * The configuation file must be specified in the sdf within the plugin element within the yarpConfigurationFile tags
+ * The configuration file must contain two groups [WRAPPER] and [DRIVER]
+ * 
+ * Parameters accepted in the config argument of the open method for [WRAPPER]:
+ * 
+ * | Parameter name| Type |Units|Default Value|Required |                 Description                |              Notes             |
+ * |:-------------:|:----:|:---:|:-----------:|:-------:|:------------------------------------------:|:------------------------------:|
+ * | name          |string|  -  |    -        |   Yes   | full name of the port opened by the device | MUST start with a '/' character|
+ * | period        |double|  ms |    20       |   No    | refresh period of broadcast value in ms    |    optional, default 20ms      |   
+ * | device        |string|  -  |    -        |   Yes   | name of the network wrapper to connect to  | MUST be set to analogServer    |    
+ *
+ * Parameters accepted in the config argument of the open method for [DRIVER]:
+ * 
+ * | Parameter name|      Type     |Units|Default Value|Required |                       Description                      |                        Notes                     |
+ * |:-------------:|:-------------:|:---:|:-----------:|:-------:|:------------------------------------------------------:|:------------------------------------------------:|
+ * | device        |    string     |  -  |    -        |   Yes   | name of the device driver to instantiate               | MUST be set to gazebo_contactloadcellarray|
+ * | loadCellNames |list of strings|  -  |    -        |   Yes   | list of load cell names, eg. (lr rr lf rf)             | locations MUST be obtained manually from urdf/sdf|
+ * | loadCellX     |list of doubles|  m  |    -        |   Yes   | list of load cell location X coordinate wrt link origin| MUST be the same size and order as loadCellName  |
+ * | loadCellY     |list of doubles|  m  |    -        |   Yes   | list of load cell location Y coordinate wrt link origin| MUST be the same size and order as loadCellName  |
+ * | loadCellZ     |list of doubles|  m  |    -        |   Yes   | list of load cell location Z coordinate wrt link origin| MUST be the same size and order as loadCellName  |
+ * | linkName      |     string    |  -  |    -        |   Yes   | name of the link associated to the contact sensor      |                            -                     |
+ *
+ * NOTE: Although being sensor-related, this plugin is implemented as a model plugin instead of a sensor plugin. This design choice is due to the fact that transformations related
+ * to the link were required for computations in this device; which could not be achieved through a sensor plugin, since Gazebo inherently divides its
+ * physics engine and sensor engine in its software architecture. As a model plugin, the link information can be directly accessed,
+ * however the sensor name and collision name related to the link are obtained by parsing through the SDF directly. Again, this design decision
+ * of parsing through the sdf is made due to the fact that the Link information does not give the type of sensors but only sensor counts.
+ * 
+ * This plugin in its current version assumes a only single collision element is associated to the specified link
+ * This sensor plugin in its current version handles only one link with a plugin instance 
+ * (i.e., multiple links cannot be specified per plugin, but multiple plugins can be instantiated through the sdf)
+ * The collision element in the contact sensor added to the specific link in the sdf should be set to {link_name}_collision
+ * The name of the collision element associated to the link should be set to default (i.e. "name" parameter should not be set so that it can take
+ * the default name {link_name}_collision. Otherwise, things become worse).
+ * 
+ * WARNING: Please note that the current version of this plugin is not tested on (heavily) nested models and also rather does not support the models
+ * having links with non-default collision element names.
+ * Refer discussion in \url https://github.com/robotology/gazebo-yarp-plugins/pull/338/files/e757d8a40d21eebf8c55db9e1c47bf397a134c43#r161473858
+ * for further information.
+ */
+class yarp::dev::GazeboYarpContactLoadCellArrayDriver : public yarp::dev::IAnalogSensor,
+                                                               public yarp::dev::DeviceDriver,
+                                                               public yarp::dev::IPreciselyTimed
+    {
+        public:
+        GazeboYarpContactLoadCellArrayDriver();
+        virtual ~GazeboYarpContactLoadCellArrayDriver();
+
+        /**
+         * Update callback from Gazebo
+         * Gets the contact information from the link contact sensor
+         * Computes equivalent wrench at the link CG (location CG, rotation link origin)
+         * Transforms this wrench to the link origin (same force, moment is transformed)
+         * Maps the transformed wrench to contact normal forces at load cell locations
+         */
+        void onUpdate(const gazebo::common::UpdateInfo& /*_info*/);
+    
+        // Device Driver
+        virtual bool open(yarp::os::Searchable& config);
+        virtual bool close();
+    
+        // Analog sensor
+        virtual int read(yarp::sig::Vector &out);
+        
+        // Analog sensor unimplemented - not required in simulation
+        virtual int getState(int ch);
+        virtual int getChannels();
+        virtual int calibrateChannel(int ch);
+        virtual int calibrateChannel(int ch, double v);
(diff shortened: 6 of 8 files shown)


plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
```diff
@@ -0,0 +1,205 @@
...
+ * however the sensor name and collision name related to the link are obtained by parsing through the SDF directly. Again, this design decision
+ * of parsing through the sdf was made due to the fact that the Link information does not give the type of sensors but only sensor counts.
+ * 
+ * This plugin in its current version assumes a only single collision element is associated to the specified link
+ * This sensor plugin in its current version handles only one link with a plugin instance 
+ * (i.e., multiple links cannot be specified per plugin, but multiple plugins can be instantiated through the sdf)
+ * The collision element in the contact sensor added to the specific link in the sdf should be set to {link_name}_collision
+ * The name of the collision element associated to the link should be set to default (i.e. "name" parameter should not be set so that it can take
+ * the default name {link_name}_collision. Otherwise, things become worse.)
+ */
```
traversaro: 😍

prashanthr05: What's this excitement about ? ;D

prashanthr05:
The code is compatible and tested with gazebo 7, however some functions of gazebo 7 libraries are either deprecated or not supported in gazebo 8. I will make the necessary changes and push soon.

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
```diff
@@ -0,0 +1,205 @@
...
+ * gazebo ContactSensor instance. Finds an equivalent wrench acting at the CG of the specified link. Transforms this
+ * equivalent wrench to the origin of the specified link. This transformation is necessary since the load cell locations are
+ * with respect to the link origin. Finally, the transformed equivalent wrench is mapped to an arrary of contact normal forces acting
+ * at the load cell locations. 
+ * 
+ * This mapping problem is an infinite solution problem and is solved by obtaining a pseudo inverse, 
+ * which gives the results with a minimized norm.
+ * 
+ * The configuation file must be specified in the sdf within the plugin element within the yarpConfigurationFile tags
+ * The configuration file must contain two groups [WRAPPER] and [DEVICE]
```
prashanthr05: This should be changed to [DRIVER] instead of [DEVICE]

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArray.cc
```diff
@@ -69,12 +69,7 @@ namespace gazebo
         // Insert the pointer in the singleton handler for retrieving it in the yarp driver
         GazeboYarpPlugins::Handler::getHandler()->setRobot(boost::get_pointer(_parent));
 
-#if GAZEBO_MAJOR_VERSION >= 7
-        m_gazeboRobotName = _parent->GetScopedName();
-#else
-        m_gazeboRobotName = _parent->ScopedName();
-#endif
-    
+        m_gazeboRobotName = _parent->GetName();
```
traversaro: You went from scoped name to name.. they are the same?

prashanthr05: I don't know why I made this implusive decision. Reverting back to GetScopedName() since this method is supported through gazebo8 as well

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
traversaro: I love good documentation!

plugins/contactpressuresensorsarray/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+# Copyright (C) 2006 Istituto Italiano di Tecnologia ADVR & iCub Facility & RBCS Department
```
traversaro: Update the copyright year to 2018

plugins/contactpressuresensorsarray/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+# Authors: see AUTHORS file.
```
traversaro: Please update the AUTHORS list, thanks!

plugins/contactpressuresensorsarray/include/gazebo/ContactPressureSensorsArray.hh
```diff
@@ -0,0 +1,86 @@
...
+
+#include <sdf/sdf_config.h>
+#include <sdf/Element.hh>
+#include <sdf/Param.hh>
+
+
+namespace sdf 
+{
+    class Element;
+}
```
traversaro: If you include the `sdf/Element.hh` header, there is no need to forward declare the `sdf::Element` class.

plugins/contactpressuresensorsarray/include/gazebo/ContactPressureSensorsArray.hh
```diff
@@ -0,0 +1,86 @@
...
+    
+       /**
+        * Saves the gazebo pointer, creates the device driver
+        */
+        void Load(physics::ModelPtr _parent, sdf::ElementPtr _sdf);
+    
+        private:
+       
+       /**
+        * Polydriver instance to connect to analaog sensor server
```
traversaro: analaog typo

plugins/contactpressuresensorsarray/include/gazebo/ContactPressureSensorsArray.hh
```diff
@@ -0,0 +1,86 @@
...
+        */
+        yarp::dev::PolyDriver m_dev_wrapper;
```
traversaro: Not a big problem, but I would keep the naming style of the attributes consistent: either `m_devWrapper` or `m_dev_wrapper` , without changing across different attributes.

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
```diff
@@ -0,0 +1,216 @@
...
+    {
+        class GazeboYarpContactPressureSensorsArrayDriver;
+    }
+}
+
+/**
+ * YARP Device Driver for an array of contact pressure sensors
+ * 
+ * This device is used to simulate the contact normal forces that might be 
+ * observed by an arbitrary number of load cells at specific locations with respect
```
traversaro: "that might be observed" --> measured

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
```diff
@@ -0,0 +1,216 @@
...
+ * to a given link's origin frame. It uses the instance of a Gazebo Contact Sensor which
+ * provides contact information of the link (specifically providing wrenches acting at the CG).
```
traversaro: Define "wrenches" and "CG" before using them.

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * YARP Device Driver for an array of contact pressure sensors
+ * 
+ * This device is used to simulate the contact normal forces that might be 
+ * observed by an arbitrary number of load cells at specific locations with respect
+ * to a given link's origin frame. It uses the instance of a Gazebo Contact Sensor which
+ * provides contact information of the link (specifically providing wrenches acting at the CG). 
+ * This information is transformed and mapped to the desired contact normal forces.
+ * 
+ * In its current version, this device reads the location of the load cells (to be mentioned manually from the urdf or sdf)
```
traversaro: "to be mentioned" ?

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
```diff
@@ -0,0 +1,216 @@
...
+ * with respect to the link origin. Finally, the transformed equivalent wrench is mapped to an arrary of contact normal forces acting
+ * at the load cell locations. 
+ * 
+ * This mapping problem is an infinite solution problem and is solved by obtaining a pseudo inverse, 
+ * which gives the results with a minimized norm.
+ * 
+ * The configuation file must be specified in the sdf within the plugin element within the yarpConfigurationFile tags
+ * The configuration file must contain two groups [WRAPPER] and [DRIVER]
+ * Parameters accepted in the config argument of the open method for [WRAPPER]:
+ * | Parameter name| Type |Units|Default Value|Required |                 Description                |              Notes             |
```
traversaro: I am afraid we need a space between the line `Parameters accepted in the config argument of the open method for [WRAPPER]:` and `| Parameter name| Type |Units|Default Value|Required |                 Description                |              Notes             |`, otherwise the markdown table will not be correctly generated.

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
```diff
@@ -0,0 +1,216 @@
...
+ * observed by an arbitrary number of load cells at specific locations with respect
+ * to a given link's origin frame. It uses the instance of a Gazebo Contact Sensor which
+ * provides contact information of the link (specifically providing wrenches acting at the CG). 
+ * This information is transformed and mapped to the desired contact normal forces.
+ * 
+ * In its current version, this device reads the location of the load cells (to be mentioned manually from the urdf or sdf)
+ * from a configuration file. Reads the contact forces and torques acting at the gazebo collision level through a 
+ * gazebo ContactSensor instance. Finds an equivalent wrench acting at the CG of the specified link. Transforms this
+ * equivalent wrench to the origin of the specified link. This transformation is necessary since the load cell locations are
+ * with respect to the link origin. Finally, the transformed equivalent wrench is mapped to an arrary of contact normal forces acting
```
traversaro: "arrary" typo

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
```diff
@@ -0,0 +1,216 @@
...
+ * | device        |    string     |  -  |    -        |   Yes   | name of the device driver to instantiate               | MUST be set to gazebo_contactpressuresensorsarray|
+ * | loadCellNames |list of strings|  -  |    -        |   Yes   | list of load cell names, eg. (lr rr lf rf)             | locations MUST be obtained manually from urdf/sdf|
+ * | loadCellX     |list of doubles|  m  |    -        |   Yes   | list of load cell location X coordinate wrt link origin| MUST be the same size and order as loadCellName  |
+ * | loadCellY     |list of doubles|  m  |    -        |   Yes   | list of load cell location Y coordinate wrt link origin| MUST be the same size and order as loadCellName  |
+ * | loadCellZ     |list of doubles|  m  |    -        |   Yes   | list of load cell location Z coordinate wrt link origin| MUST be the same size and order as loadCellName  |
+ * | linkName      |     string    |  -  |    -        |   Yes   | name of the link associated to the contact sensor      |                            -                     |
+ *
+ * Implemented as a model plugin instead of a sensor plugin. This design choice was due to the fact that transformations related
```
traversaro: The phrase  "Implemented as a model plugin instead of a sensor plugin. " is a bit abrupt. I will change it in "Note: this plugin was implemented as a model plugin instead of a sensor plugin." or something similar.

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
```diff
@@ -0,0 +1,216 @@
...
+        bool initContactSensor();
+        
+       /**
+        * Reads the configuration file for load cell locations
+        * \return true if successful
+        */
+        bool configure(yarp::os::Property &pluginParameters);
+        
+       /**
+        * Prepares a matrix as a funciton of loadcell locations
```
traversaro: typo "funciton"

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
prashanthr05: Likewise! :D

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArray.cc
```diff
@@ -0,0 +1,193 @@
...
+    
+        driver_list.push(&m_dev_driver, "dummy");
+    
+        if (!m_imultwrapper->attachAll(driver_list))
+        {
+            yError() << "GazeboYarpContactPressureSensorsArray Plugin failed: Error in connecting the wrapper and device";
+        }
+     
+        yInfo() << "GazeboYarpContactPressureSensorsArray: Opened device driver and attached to Analog Server wrapper \n";
+    } // end Load(sensors::SensorPtr _parent, sdf::ElementPtr _sdf)
```
traversaro: Note: I am not a big fan of this "end of scope" comments, as it is difficult to maintain them when you change names of functions. However if you like them we can keep them.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+
+GazeboYarpContactPressureSensorsArrayDriver::~GazeboYarpContactPressureSensorsArrayDriver()
+{
+
+}
+
+void GazeboYarpContactPressureSensorsArrayDriver::onUpdate(const gazebo::common::UpdateInfo& _info)
+{ 
+    gazebo::msgs::Contacts contacts;
+    contacts = this->m_sensor->Contacts();
```
traversaro: I guess this involves a copy of quite a big object. Have you considered using a const reference for this?

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+    ignition::math::Vector3d resultantForceonCG;
+    ignition::math::Vector3d resultantMomentonCG;
+        
+    double timestamp = -1;
+    for (size_t i = 0; i < contacts.contact_size(); i++)
+    {
+        gazebo::common::Time currentStamp;
+        for (int j=0; j < contacts.contact(i).position_size(); j++)
+        {
+            double t = contacts.contact(i).time().sec() + TIME_CONVERSION_NS*contacts.contact(i).time().nsec();
```
traversaro: I think this line can be simplified to:
~~~
double t = contacts.contact(i).time.Double() 
~~~

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+#include <yarp/os/LockGuard.h>
+#include <yarp/os/Log.h>
+#include <yarp/os/LogStream.h>
+#include <yarp/math/Math.h>
+#include <yarp/math/SVD.h>
+
+#include <sdf/Param.hh>
+
+#define DEBUG 0
+#define N_AXIS 3
```
traversaro: This is used only in `GazeboYarpContactPressureSensorsArrayDriver::prepareMappingMatrix()`, so I would transform it in a local const variable of that method. In general it is better to avoid to use the preprocessor unless it is strictly necessary.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+#define TIME_CONVERSION_NS 1.0e-9
```
traversaro: Based on other comments, this should be removed.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+            if (timestamp < 0)
+            {
+                timestamp = t;
+            }
+
+            // This block allows addition of wrench components only if they arrive in same timestamp
+            if (!ignition::math::equal(t, timestamp))
+                continue;
+
+            gazebo::msgs::JointWrench wrench = contacts.contact(i).wrench(j);
```
traversaro: Use a reference here?

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+{
+
+}
+
+void GazeboYarpContactPressureSensorsArrayDriver::onUpdate(const gazebo::common::UpdateInfo& _info)
+{ 
+    gazebo::msgs::Contacts contacts;
+    contacts = this->m_sensor->Contacts();
+    
+    ignition::math::Vector3d resultantForceonCG;
```
traversaro: Are you sure that this vectors are initialized to zero? From the context it seems that you are assuming this, but this is not documented. If this is actual the case ( https://bitbucket.org/ignitionrobotics/ign-math/src/5bf0933cc7325f6ccb149676a857bb725b163d35/include/ignition/math/Vector3.hh?at=default&fileviewer=file-view-default#Vector3.hh-58 ) it is a non documented behavior, so eventually they can change it. I would recommended to explicitly set the vector to zero to be future-safe.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+    // Compute the pseudo-inverse to determine the mapping matrix
+    this->m_mapWrenchtoNormalForce = yarp::math::pinv(A);
+
+    return true;
+}
+
+// gazebo::math::Quaternion method GetAsMatrix() is deprecated from version 8.0
+// and ignition::math::Quaternion does not support this method, so porting it
+// from gazebo::math just to support double type values, change if necessary
+ignition::math::Matrix3d GetQuaternionAsMatrix3d(ignition::math::Quaterniond q)
```
traversaro: I guess you can use the https://osrf-distributions.s3.amazonaws.com/ign-math/api/1.0.0/classignition_1_1math_1_1Matrix3.html#a03d18617b019eb137c318bd60d653794 Quaternion-based Matrix3d constructor for this.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+
+int GazeboYarpContactPressureSensorsArrayDriver::read(yarp::sig::Vector& out)
+{
+    yarp::os::LockGuard guard(m_dataMutex);
+  
+    // send a zero vector in case the data is not computed already
+    if (!m_dataAvailable)
+    {
+        // Introducing average time delay for the output data to be ready
+        usleep(9000);
```
traversaro: This is evil. 
If no data is available, simply return `AS_ERROR` .

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
prashanthr05: copy the load cell locations from the urdf/sdf into the config file ?

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArray.cc
prashanthr05: I had just put it there, since it was a really big function block... I will remove it!

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
prashanthr05: Actually I hadn't put enough thought into the memory perspective... Thanks for opening my eyes. I think this could help improve the was-failing RTF.

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
traversaro: If you write the full sentence you want to add in the documentation it is easier for me to comment on it.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArray.cc
```diff
@@ -0,0 +1,193 @@
...
+        }
+    
+        driver_list.push(&m_dev_driver, "dummy");
+    
+        if (!m_imultwrapper->attachAll(driver_list))
+        {
+            yError() << "GazeboYarpContactPressureSensorsArray Plugin failed: Error in connecting the wrapper and device";
+        }
+     
+        yInfo() << "GazeboYarpContactPressureSensorsArray: Opened device driver and attached to Analog Server wrapper \n";
```
traversaro: I think this can be removed.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+    A.resize(N_AXIS ,n_cells);
+  
+    // Add the respective columns to the mapping matrix
+    for (int i = 0; i < n_cells; i++)
+    {
+        yarp::sig::Vector columnMap;
+        columnMap.clear();
+        columnMap.push_back(1);
+        columnMap.push_back(this->m_loadCellLocations[i][1]); // r_y : y-intercept of the contact point with respect to link origin
+        columnMap.push_back(-this->m_loadCellLocations[i][0]); // -r_x : x-intercept of the contact point with respect to link origin
```
traversaro: I think here we are assuming that the load cells are distributed over a plane perpendicular to the z-axis of the link frame, and that the load cells are measuring the force over the z direction. This assumptions should be documented in the plugin docs.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+    // uncomment the block if you think this assumption is invalid  
+    /* while(!mgr->SensorsIntialized())
+     * {
+     *     yTrace() << "GazeboYarpContactPressureSensorsArrayDriver: Waiting for sensors to initialize";
+     *     usleep(10000);
+     * }
+     */
+  
+    gazebo::sensors::SensorPtr contact = mgr->GetSensor(sensorName); 
+    this->m_sensor = dynamic_cast<gazebo::sensors::ContactSensor*>(contact.get());
```
traversaro: Check if the dynamic cast was successful and fail otherwise.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+     */
+  
+  
+    this->m_linkCollisionName = m_robot->GetName() + "::" + m_sensorLink->GetName() + "::" + collisionName;
+  
+    yarp::os::LockGuard guard(m_dataMutex);
+    gazebo::sensors::SensorManager *mgr = gazebo::sensors::SensorManager::Instance();
+  
+    // Assuming the sensors are initialized 
+    // uncomment the block if you think this assumption is invalid
```
traversaro: Have you assessed that? If this is the case, we can remove the commented code.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+    // Get collision name 
+    sdf::ElementPtr collision = (sensorsdf.get()->GetElement("contact")).get()->GetElement("collision"); 
+    std::string collisionName = collision.get()->GetValue()->GetAsString();
+    /* std::string tmp = m_sensorLink->GetName() + "::" + collisionName + "::" + sensorName;
+     * sensorName.clear();
+     * sensorName = tmp;
+     */
+  
+  
+    this->m_linkCollisionName = m_robot->GetName() + "::" + m_sensorLink->GetName() + "::" + collisionName;
```
traversaro: It would be worth to check if this works by trying it on a nested model, such as "iCub Fixed".

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
traversaro: You can also transform it in a "If the sensor are not initialized, fail".

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
prashanthr05: For some reason, `time` is pointing towards `gazebo::msgs::Time` instead of `gazebo::common::Time` and makes it unable to access the `Double` method of the `Time` class. So, sticking to the lengthy version.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+        }
+    }
+
+#if DEBUG 
+     std::cout << "Wrench at CG [ " << resultantForceonCG.X() << " " <<  resultantForceonCG.Y() << " " << resultantForceonCG.Z() << " " << resultantMomentonCG.X() << " " <<  resultantMomentonCG.Y() << " " << resultantMomentonCG.Z() << " ]" << std::endl;
+#endif
+     
+    // Transform the wrench applied on the Link CG frame to the Link Origin Frame
+    // consider L: Link Origin Frame, C: Link CG Frame, ^: cross product operator givng a skew symmetric matrix
+    // L_f = L_R_C * C_f
```
traversaro: According to https://bitbucket.org/osrf/gazebo/issues/545/request-change-contact-reference-frame#comment-17039155 , the force and torques are already expressed with the orientation of the link frame (L). Are you sure that instead they are expressed in the orientation of the inertia matrix (C). Or perhaps in the link that we are testing L_R_C is the identity and for this reason everything works fine?

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
prashanthr05: Thanks, great!

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
prashanthr05:
```
    if (!m_dataAvailable)
    {
        // Introducing average time delay for the output data to be ready
        usleep(9000);
        return AS_TIMEOUT;
    }
```

Is this block ok ? I'm considering the `usleep(9000)`, because the first measurement after all the computations becomes ready by an average time of 6ms, and during that time, this `read` method floods the output with timeout prints. If you think this is not necessary, I will remove the usleep as well.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
prashanthr05: #if DEBUG 0 stays or leaves ?

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+
+#if DEBUG  
+    std::cout << " Absolute pose [ " << W_H_L.Pos().X() << " " <<  W_H_L.Pos().Y() << " " << W_H_L.Pos().Z() << " ]"<< std::endl;
+#endif
+  
+    // To find L_R_C and L_o_C
+    // L_R_C = L_R_W * W_R_C
+    // L_o_C = L_R_W (W_o_C - W_o_L)
+    ignition::math::Matrix3d L_R_W = GetQuaternionAsMatrix3d(W_H_C_rotL.Rot().Inverse());
+    this->m_linkOrigin_Rot_linkCG = L_R_W*(GetQuaternionAsMatrix3d(W_H_C.Rot()));
```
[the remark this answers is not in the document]
prashanthr05: I had not gone through the `Inertial` API before. I just noticed there is a Pose related method in it.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
prashanthr05: Apparently, with the contact sensor plugin, the real time factor drops down to around 0.12 and without the plugin, real time factor stays around 0.69

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
prashanthr05: http://osrf-distributions.s3.amazonaws.com/gazebo/api/dev/classgazebo_1_1sensors_1_1ContactSensor.html#a5e8c0561a71708862c4304d6174512e9

Here it says, The forces and torques are applied at the CG of perspective links for each collision body, specified in the inertial frame. That is why I took was using the transformation.

And yes, actually L_R_C in the current test case is identity.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
traversaro: What is the period of the analog wrapper? Even if it is 1 ms, there should be just ~6 messages printed due to this. Hard-coded delays are typically problematic especially in complex systems, and personally I prefer to avoid them as much as possible.

prashanthr05: I understand the rationale. The period was set to a very low value. I agree that it's better to play with higher level parameters rather than introducing hard coded delays. Thank you!

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
traversaro: Upon a quick check, I can conclude two things: 
* In that documentation "inertial frame" was meant as "world frame", i.e. inertial frame according to Newton's first law, not "frame in which the inertia is expressed". 
* The documentation was originally added in 2013, but in 2015 the behavior of the function changed and the documentation was not updated.  

Details: 
* The documentation was originally added in https://bitbucket.org/osrf/gazebo/pull-requests/313/add-documentation-describing-data-returned (2013). At the same time the author of the PR opened an issue to return the data in "local frame": https://bitbucket.org/osrf/gazebo/issues/540/update-contactsensor-for-getting-contact . This probably means that at that time the contact data was not in a local frame. 
* The behavior of the function was changed in PR https://bitbucket.org/osrf/gazebo/pull-requests/355/bullet-contact-sensor (2015), but the documentation was never updated. 

I think it is safe to assume that those quantities are expressed with the orientation of the link frame. @prashanthr05 can you open an issue about this in the Gazebo bugtracker? Feel free to copy&paste part of this comment in the issue.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
prashanthr05: I noticed in iCub Fixed... the collision name is set {link_name}_collision. In this case the collision name is actually set to {link_name}_fixed_joint_lump_{link_name}_collision_collision, according to the observations made in http://answers.gazebosim.org/question/15095/gazebo-contact-sensor-added-to-a-robot-link-in-urdf-file-is-not-working-as-expected/

This is not handled by the plugin. However, need to test the plugin for heavily nested models.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
traversaro: I see, ok.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
```diff
@@ -0,0 +1,516 @@
...
+    out.resize(m_contactNormalForces.size());
+    out = m_contactNormalForces;
+  
+    return AS_OK;
+}
+
+
+yarp::os::Stamp GazeboYarpContactPressureSensorsArrayDriver::getLastInputStamp()
+{
+    return this->m_stamp;
```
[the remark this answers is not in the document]
traversaro: Ok, now it is updated.

plugins/contactpressuresensorsarray/include/yarp/dev/ContactPressureSensorsArrayDriver.h
```diff
@@ -0,0 +1,225 @@
...
+ * physics engine and sensor engine in its software architecture. As a model plugin, the link information can be directly accessed,
+ * however the sensor name and collision name related to the link are obtained by parsing through the SDF directly. Again, this design decision
+ * of parsing through the sdf is made due to the fact that the Link information does not give the type of sensors but only sensor counts.
+ * 
+ * This plugin in its current version assumes a only single collision element is associated to the specified link
+ * This sensor plugin in its current version handles only one link with a plugin instance 
+ * (i.e., multiple links cannot be specified per plugin, but multiple plugins can be instantiated through the sdf)
+ * The collision element in the contact sensor added to the specific link in the sdf should be set to {link_name}_collision
+ * The name of the collision element associated to the link should be set to default (i.e. "name" parameter should not be set so that it can take
+ * the default name {link_name}_collision. Otherwise, things become worse.)
```
traversaro: Not a native speaker, but I don't think you can close the a period inside a bracket.

prashanthr05: Noted. Will push this change along with some major commit.

plugins/contactpressuresensorsarray/src/ContactPressureSensorsArrayDriver.cpp
prashanthr05: I think we have implemented only a corner case. What to do with this kind of generic handling ?

traversaro:
I think there are just a few docs issue ( https://github.com/robotology/gazebo-yarp-plugins/pull/338#discussion_r161703931 , the fact that nested models are not supported should be documented) and then we can merge. @prashanthr05 can you rebase the commits in a single commit? See https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history for more info on this.

prashanthr05:
Noted. Will add that drawback in the documentation. And rebase the commits as a single commit.

traversaro:
Sorry, another request: it would be great if you could update the table in https://github.com/robotology/gazebo-yarp-plugins/blob/master/doc/embed_plugins.md#plugins-exposing-functionalities-available-on-real-robots .

traversaro:
Thanks!

Issue #337, opened 2018-01-11
Access to state of links and models in Gazebo


xEnVrE:
It seems that none of the plugins available in `gazebo-yarp-plugins` allows to access to the state of the links and models loaded inside Gazebo. While for links one could have developed its own plugin providing the state of the links of a robot it could be useful to have access to the state of the models loaded inside Gazebo. This way one could get the position and the orientation of objects that the robot is interacting with and, for example, compare it with an estimate provided by a localization algorithm.

This feature is implemented in `gazebo_ros` for example ([link](http://gazebosim.org/tutorials/?tut=ros_comm#GazeboPublishedTopics)) with topics providing the state of all the links and all the models.

Among the plugins available in `gazebo-yarp-plugins` the plugin `gazebo_yarp_worldinterface` offers a __Thrift__ based API that allows to create standard objects, move them, get their pose and so on. Could it be an option to add a method that allows the user to access to the state of a model loaded inside Gazebo but not created using the plugin, e.g. loaded using a `.sdf` file?

Since sending the state of all the models on a port seems an overkill other solutions may be proposed:

- a method like
```
Pose getModelPose(model_name)
```
where `model_name` is the name of the model within Gazebo, returning the current pose of the model 
- a method like
```
bool requestModelPose(model_name, port_name, send_rate)
```
After calling this method the plugin should start sending the state of the model with name `model_name` over the port `port_name` with a given rate `send_rate`.

It would be great to know your opinions about this!

traversaro:
I like the idea, and I think the use case is clear. 
Some brainstorming (don't take this comment as gospel please : ) ) :
Some comments on the proposed design: 
* You are proposing to add a method to open a port that publishes data. Do you think you should also have a method to close the port? 
* The `gazebo_yarp_worldinterface` plugin is already quite complicated, so perhaps it could make sense to implement this in a separate plugin. What to do think?

Some possible alternative designs to achieve the same use case:
* Instead of publishing each model information in its own port, you could publish all the model poses in one port/topic, similar to what the ros plugins does. Drawbacks: in complex simulation scenario, the amount of data published on that port could be big. 
[...]

PR #340, opened 2018-01-26
Model Pose Publisher model plugin

xEnVrE:
This is a possible implementation of a Model plugin that allows to retrieve the absolute pose of a model inserted within the Gazebo environment, as discussed in #337.

It uses the base class `gazebo::ModelPlugin`  to access the model and obtain its pose defined as the absolute pose, with respect to the Gazebo world frame, of the root link of the model 
```
<sdf>
  <model name="object">
    <!-- pose of model frame w.r.t Gazebo world -->
    <pose> x_m y_m z_m roll_m pich_m yaw_m </pose>

    <link name="object_root_link">
      <!-- pose of the root link w.r.t the model frame -->
      <pose> x_l y_l z_l roll_l pich_l yaw_l </pose>    
      ...
    </link>
    
    <plugin name='...' filename='libgazebo_yarp_modelPosePublisher.so'>
      <period>1.0</period>
    </plugin>
  </model>
<\sdf>
```
This pose depends on both the transformation between the Gazebo world frame and the model frame and the transformation between the model frame and the root link of the model.

The pose is published within the plugin using a `yarp::dev::FrameTransformClient` that connects to a `yarp::dev::FrameTransformServer`. The server can be activated from the command line using `yarpdev` as explained in the documentation. The user can access to the pose of the model using a `yarp::dev::FrameTransformClient`.
__Documentation__ of the plugin and __example code__ for the final user is provided in the documentation string of the main class `GazeboYarpModelPosePublisher` available through Doxygen. Also a __tutorial__ is available in `gazebo-yarp-plugins/tutorial/model/model_state_publisher`. 

The plugin does not support Gazebo [nested models](http://gazebosim.org/tutorials?tut=nested_model&cat=build_robot) since the method `gazebo::physics::Model::WorldPose`, required to obtain the absolute pose of a model, is not working properly with nested models.

The plugin was tested with Gazebo 7 and deprecated methods in Gazebo 8 are handled in the code.

### DIFF
--- a/plugins/modelposepublisher/CMakeLists.txt
+++ b/plugins/modelposepublisher/CMakeLists.txt
@@ -1,0 +1,23 @@
+# Copyright (C) 2018 Istituto Italiano di Tecnologia 
+# Authors: see AUTHORS file.
+# CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
+
+cmake_minimum_required(VERSION 2.8.7)
+
+PROJECT(Plugin_ModelPosePublisher)
+
+include(AddGazeboYarpPluginTarget)
+
+set(LIB_COMMON_NAME gazebo_yarp_lib_common)
+if(CMAKE_VERSION VERSION_LESS 3.0.0)
+    get_property(GAZEBO_YARP_COMMON_HEADERS GLOBAL PROPERTY GAZEBO_YARP_COMMON_HEADERS)
+    unset(LIB_COMMON_NAME)
+endif()
+
+add_gazebo_yarp_plugin_target(LIBRARY_NAME modelPosePublisher
+                              INCLUDE_DIRS include/gazebo
+                              SYSTEM_INCLUDE_DIRS ${GAZEBO_YARP_COMMON_HEADERS} ${YARP_INCLUDE_DIRS} ${Boost_INCLUDE_DIRS} ${GAZEBO_INCLUDE_DIRS}  ${SDFORMAT_INCLUDE_DIRS} ${PROTOBUF_INCLUDE_DIRS} ${OGRE_INCLUDE_DIRS}
+                              LINKED_LIBRARIES ${LIB_COMMON_NAME} ${YARP_LIBRARIES} ${GAZEBO_LIBRARIES} ${Boost_LIBRARIES}
+                              HEADERS include/gazebo/ModelPosePublisher.hh
+                              SOURCES src/ModelPosePublisher.cc
+                              )
--- a/plugins/modelposepublisher/CMakeLists.txt
+++ b/plugins/modelposepublisher/CMakeLists.txt
@@ -14,7 +14,7 @@ if(CMAKE_VERSION VERSION_LESS 3.0.0)
     unset(LIB_COMMON_NAME)
 endif()
 
-add_gazebo_yarp_plugin_target(LIBRARY_NAME modelPosePublisher
+add_gazebo_yarp_plugin_target(LIBRARY_NAME modelposepublisher
                               INCLUDE_DIRS include/gazebo
                               SYSTEM_INCLUDE_DIRS ${GAZEBO_YARP_COMMON_HEADERS} ${YARP_INCLUDE_DIRS} ${Boost_INCLUDE_DIRS} ${GAZEBO_INCLUDE_DIRS}  ${SDFORMAT_INCLUDE_DIRS} ${PROTOBUF_INCLUDE_DIRS} ${OGRE_INCLUDE_DIRS}
                               LINKED_LIBRARIES ${LIB_COMMON_NAME} ${YARP_LIBRARIES} ${GAZEBO_LIBRARIES} ${Boost_LIBRARIES}
--- a/plugins/modelposepublisher/include/gazebo/ModelPosePublisher.hh
+++ b/plugins/modelposepublisher/include/gazebo/ModelPosePublisher.hh
@@ -56,7 +56,7 @@ namespace gazebo
     ///     ...
     ///     </link>
     ///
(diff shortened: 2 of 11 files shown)


traversaro:
Restarted the Travis jobs because there was some dockerhub problem.

plugins/modelposepublisher/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+# Copyright (C) 2018 Istituto Italiano di Tecnologia 
+# Authors: see AUTHORS file.
```
traversaro: Please add yourself to the AUTHORS file, thanks!

traversaro:
> The plugin does not support Gazebo nested models since the method gazebo::physics::Model::WorldPose, required to obtain the absolute pose of a model, is not working properly with nested models. 

Interesting. There is an issue on the Gazebo bug tracker on this?

traversaro:
cc @prashanthr05

prashanthr05:
cc @randaz81

Yeshasvitvs:
I wrote a small plugin called [accesslinkpose](https://github.com/Yeshasvitvs/gazebo-yarp-plugins/tree/topology_estimation_plugins/plugins/accesslinkpose) to send the pose of specific links given in a config file as the following example

```
[include "gazebo_1r_2link_robotname.ini"]

port_name /${gazeboYarpPluginsRobotName}/pose_output:o
number_of_links 2
link_names first_link second_link
link_pose_type world world
```

This may be useful.

plugins/modelposepublisher/CMakeLists.txt
xEnVrE: Done!

xEnVrE:
@traversaro as regards the bug with nested models I opened an [issue](https://bitbucket.org/osrf/gazebo/issues/2410/wrong-gazebo-physics-model-worldpose-for). 

However I think it is important to stress that [nested models](http://gazebosim.org/tutorials?tut=nested_model&cat=build_robot) are quite different from nesting using the `<include>` tag in a SDF (as happens in [icub-gazebo](https://github.com/robotology/icub-gazebo) SDFs). As stated in the  [documentation](http://gazebosim.org/tutorials?tut=nested_model&cat=build_robot) 

>The `<include>` element works by taking all the links from the included model and embedding them into the parent model

hence when using `<include>` the nested models do not exist anymore in the SDF. If one access to the pose of one the resulting links using `gazebo::physics::Link::WordPose()` then it should work.

However since nested models like

```
<sdf version="1.5">

  <model name="root_model">
    <pose> 1 1 0 0 0 0 </pose>
    <link name="root_model_root_link">
      <!-- standard sdf content here -->
    </link>

    <!-- nested model -->
    <model name="nested_model">
      <!-- pose relative to the parent model -->
      <pose> 0 0 1 0 0 0</pose>
      <link name="nested_model_root_link">
        <!-- standard sdf content here -->
      </link>

      <!-- plugin inserted inside the nested model -->
      <plugin name='plugin' filename='plugin_name.so'></plugin>
    </model>

  </model>

</sdf>
```
are also allowed, i.e. without using `<include>`, I tried to put a plugin that uses `gazebo::physics::Model::WorldPose()` directly in the nested model (like in the example above) and found that it does not work properly, e.g. the relative pose of the nested model with respect to the 
root model is not reflected in the pose of the nested model and other strange things described in the issue.

traversaro:
> @traversaro as regards the bug with nested models I opened an issue. 

Thanks, I would add a link to that issue in the documentation when you mention that nested models are not supported.
Furthermore, if you want to have fun I can show you how to can add that example as a regression test to the Gazebo test suite. : )

xEnVrE:
@traversaro 
> Thanks, I would add a link to that issue in the documentation when you mention that nested models are not supported.

I agree.

>Furthermore, if you want to have fun I can show you how to can add that example as a regression test to the Gazebo test suite. : )

It would be great if you can point me to some reference regarding this : )

plugins/modelposepublisher/include/gazebo/ModelPosePublisher.hh
```diff
@@ -107,7 +107,7 @@ namespace gazebo
     ///
     /// This plugin __does not__ support <a href="http://gazebosim.org/tutorials?tut=nested_model&cat=build_robot">Gazebo nested models</a>
     /// since the method `gazebo::physics::WorldPose()` does not work properly
-    /// with nested models.
```
traversaro: This works fine, but consider that Doxygen also support Markdown, so you can also write markdown-style links.

traversaro:
> It would be great if you can point me to some reference regarding this : ) 

http://gazebosim.org/tutorials?tut=contrib_code
http://gazebosim.org/tutorials?cat=guided_a&tut=guided_a1 

And if you need an easy way to setup Gazebo and its deps from source : https://github.com/traversaro/gazebo-superbuild .

traversaro:
Thanks @xEnVrE for the awesome plugin and @barbalberto for the review. 
I have one last request for @xEnVrE : can you add the links to the plugin documentation in the `doc/embed_plugins.md` file? Thanks!

traversaro:
Just noticed now that the plugin name is `gazebo_yarp_modelPosePublisher`, while almost all the other plugins do not use camel case in their naming. Sorry for this last request (but if we stick to a name then it would be extremely difficult to change it to something else in the future) but could you name the plugin `gazebo_yarp_modelposepublisher` ? Thanks : )

xEnVrE:
>Just noticed now that the plugin name is gazebo_yarp_modelPosePublisher, while almost all the other plugins do not use camel case in their naming. Sorry for this last request (but if we stick to a name then it would be extremely difficult to change it to something else in the future) but could you name the plugin gazebo_yarp_modelposepublisher ? Thanks : )

@traversaro  of course : )

traversaro:
Thanks, merging.

PR #247, opened 2016-06-14
[controlboard] Fix issue with ResetWorld

traversaro:
Fix https://github.com/robotology/gazebo-yarp-plugins/issues/92 .

A callback for the resetWorld event was added.
During a World Reset, the state of the controller is reset to the initial state
(including control mode,  interaction mode and state of the trajectory generators)

@jeljaik @rlober 
Can you test it?

cc @DanielePucci

Issue #92, opened 2014-04-18
Gazebo Reset World causes unrealistic joint torques


iron76:
A "Reset World" action results in strange values for joint torques very different (~1e4) from the torques that the system gives at the first insertion of the model.

traversaro:
We should add a callback on the World Reset event [1] to properly reset the controller state.
[1] https://bitbucket.org/osrf/gazebo/issue/1375/add-events-connectworldreset-and-events

traversaro:
http://answers.gazebosim.org/question/8552/how-can-i-reset-the-simulation-properly-in-gazebo/ actually we should properly implement the Reset() method in the controlboard plugin.

traversaro:
cc @barbalberto apparently we just need to write the `Reset()` method for the ControlBoard plugin .

barbalberto:
Great @traversaro, thanks. I'll check it asap

rlober:
Any news on this? Pingalingpingpingdingdong

traversaro:
I drunkly promised to fix it, so I guess now I need to actually fix this. :D

rlober:
Hahaha YES!!!!

DanielePucci:
:D

rlober:
Any news on this? @traversaro @iron76 @francesco-romano

rlober:
@traversaro If you need some help on this just point @jeljaik and myself to the right file and we can try to figure it out. This is pretty critical for us so I am happy to volunteer some time to this problem, I just need to know where to start.

traversaro:
Yes, sorry guys! 
My impression is that the Gazebo JointController is not properly resetted on World Reset. 
We then need to implemented the `Reset()` method in [`gazebo::GazeboYarpControlBoard`](https://github.com/robotology/gazebo-yarp-plugins/blob/master/plugins/controlboard/src/ControlBoard.cc) and properly reset the controlboard logic and call the reset on the Gazebo JointController... the big problem that I know see is that the JointController pointers are all contained in the [`ControlBoardDriver`](https://github.com/robotology/gazebo-yarp-plugins/blob/master/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h) class, so it may more convenient to add a callback on the [`WorldReset`](https://bitbucket.org/osrf/gazebo/issues/1375/add-events-connectworldreset-and-events) event, as we already do for the `WorldUpdate` event.

PR #345, opened 2018-02-08
Added Reset Simulation

triccyx:
Added to clock gazebo plugin the possibility to reset simulation via rpc call

### DIFF
--- a/plugins/clock/src/Clock.cc
+++ b/plugins/clock/src/Clock.cc
@@ -179,7 +179,8 @@ namespace gazebo
 
     void GazeboYarpClock::reset()
     {
-        m_world->ResetEntities(gazebo::physics::Base::MODEL);
+        //m_world->ResetEntities(gazebo::physics::Base::MODEL);
+        m_world->Reset();
     }
 
     common::Time GazeboYarpClock::getSimulationTime()
--- a/plugins/clock/src/Clock.cc
+++ b/plugins/clock/src/Clock.cc
@@ -179,7 +179,7 @@ namespace gazebo
 
     void GazeboYarpClock::reset()
     {
-        m_world->Reset();
+        m_world->ResetEntities(gazebo::physics::Base::MODEL);
     }
 
     common::Time GazeboYarpClock::getSimulationTime()
--- a/plugins/clock/src/Clock.cc
+++ b/plugins/clock/src/Clock.cc
@@ -177,7 +177,7 @@ namespace gazebo
         m_world->ResetTime();
     }
 
-    void GazeboYarpClock::reset()
+    void GazeboYarpClock::resetSimulation()
     {
         //m_world->ResetEntities(gazebo::physics::Base::MODEL);
         m_world->Reset();
--- a/plugins/clock/src/Clock.cc
+++ b/plugins/clock/src/Clock.cc
@@ -177,6 +177,11 @@ namespace gazebo
         m_world->ResetTime();
     }
 
+    void GazeboYarpClock::reset()
+    {
+        m_world->Reset();
+    }
+
     common::Time GazeboYarpClock::getSimulationTime()
     {
 #if GAZEBO_MAJOR_VERSION >= 8
--- a/plugins/clock/src/Clock.cc
+++ b/plugins/clock/src/Clock.cc
@@ -179,7 +179,6 @@ namespace gazebo
 
     void GazeboYarpClock::resetSimulation()
     {
-        //m_world->ResetEntities(gazebo::physics::Base::MODEL);
         m_world->Reset();
     }
 
--- a/plugins/clock/src/Clock.cc
+++ b/plugins/clock/src/Clock.cc
@@ -179,7 +179,7 @@ namespace gazebo
 
     void GazeboYarpClock::reset()
     {
-        m_world->ResetEntities(gazebo::physics::Base::MODEL);
+        m_world->Reset();
     }
 
     common::Time GazeboYarpClock::getSimulationTime()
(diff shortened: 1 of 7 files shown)


traversaro:
I guess you want to merge the modification in the `devel` branch, right? 
You can change the target branch of the PR directly from the GitHub web user interface.

traversaro:
Modifying this thrift file is unfortunately extremely tricky, because we have multiple copies of this file disseminated across several repository, see for example https://github.com/robotology/WB-Toolbox/commit/af1fa4106a4597254b28b4f3eaecb41261600124 . I guess @diegoferigo and @S-Dafarra are the maintainers of the repositories affected by this. 

Before merging this modification, we should make sure that the client generated with the old version of the thrift file are still compatible with the server exposed by the new version of the thrift. 

The clean solution is probably to avoid this duplication. This involves either:
* Add to `gazebo-yarp-plugins` all the CMake machinery necessary for making sure that the `thrift/clock/clock_rpc.thrift` file is installed and can be found by the system. The problem of this solution would be that all the repository that use that file will now depend on gazebo-yarp-plugins and thus will depend indirectly on Gazebo. This is definitely problematic, think for example on how this will affect Windows.
* Add this interface to YARP: if we are able to ensure that this interface is relatively simulator-independent, it would make sense to have this in YARP. On a positive note, at the moment the interface is quite independent from the underlying simulator, as long as it is a fixed time-step simulator as the "Step" functions take the number of steps as input, so it could make sense to have this in YARP.

diegoferigo:
@triccyx I remember a discussion with @francesco-romano (previous WB-Toolbox maintainer) about removing that thrift-related sources as soon as YARP would have installed them in the system. I'd go as well in this direction rather than storing it into `gazebo-yarp-plugins`.

diegoferigo:
Cross ref https://github.com/robotology/wb-toolbox/issues/62

triccyx:
What is the best way to resolve the problem? Without this merge the tests won't work

diegoferigo:
I'd like to see this thrift to end into yarp. Without spending too much time to figure out if it is or isn't good enough, since it seems already sufficiently generic, why not opening a PR and ask for its inclusion? It might be a quick process. @traversaro

traversaro:
I agree with @diegoferigo .

drdanz:
I think we need to discuss this... At the moment I wouldn't know 
* where to put the thrift interfaces
* whether we want to include the thrift only, or to generate a library as well
* In case we want to add a library, we need to decide if we want it static or shared
* if shared we need to modify the thrift generator to allow to handle the dll import/export
* Also we need to commit the generated files in the repository

I suggest to merge this PR and to have a meeting about this after YARP 3 release

triccyx:
Ok for Add ClockServer thrift file #1724, but this doesn't solve the problem. 
The client at the moment can't execute the tests.  Perhaps we can merge these modifications in a branch in master and ask the client to use that branch.

traversaro:
If clients with the old interface are still able to call the methods in the new interface (I think should be the case, is anyone is able to check this?), I don't think it would be too problematic to merge this PR.

However, I think the current method added is a bit misleading, as it is called `reset`, but it actually just reset the state of the simulation, and not the time. I think `resetSimulation` should reset both time and state of the simulation, and eventually a `resetSimulationState`should just reset the state without resetting the time.

triccyx:
Any news/solutions regarding this pull request?

traversaro:
Can you address my comments from the 4th of June? 

> However, I think the current method added is a bit misleading, as it is called reset, but it actually just reset the state of the simulation, and not the time. I think resetSimulation should reset both time and state of the simulation, and eventually a resetSimulationStateshould just reset the state without resetting the time.

triccyx:
Ok I change the method name to ResetSimulation as it resets both state and time.

triccyx:
ping

traversaro:
Ok for me, but I would prefer if @diegoferigo could check this as well.

plugins/clock/src/Clock.cc
```diff
@@ -177,6 +177,12 @@ namespace gazebo
         m_world->ResetTime();
     }
 
+    void GazeboYarpClock::resetSimulation()
+    {
+        //m_world->ResetEntities(gazebo::physics::Base::MODEL);
```
traversaro:
```suggestion
```
I think we can remove this.

triccyx:
@diegoferigo friendly ping

traversaro:
> Can you confirm this?

Exactly. As you can check in the generated code for the server https://github.com/robotology/gazebo-yarp-plugins/pull/345/files#diff-9e60d5f99e3cf3eb4705af443b70c244R395 , nothing in the code changes for methods different from `resetSimulation`. 
The only possible incompatibility is related to some new software that tries to call `resetSimulation` on an old version of `gazebo-yarp-plugins`.

>     Would the signal to reset the world be propagated to all the other plugins? E.g. if there are controllers, their state should be reset too (I think yes, just checking).

Plugins can handle the reset even by attaching a callback to the `WorldReset` event, see how the controlboard plugin handle this:
* https://github.com/robotology/gazebo-yarp-plugins/blob/01ad1d1252b0827033fd57058f066cc790b893c6/plugins/controlboard/src/ControlBoardDriver.cpp#L342
* https://github.com/robotology/gazebo-yarp-plugins/blob/01ad1d1252b0827033fd57058f066cc790b893c6/plugins/controlboard/src/ControlBoardDriver.cpp#L624
old related issue and PR : https://github.com/robotology/gazebo-yarp-plugins/issues/92 and https://github.com/robotology/gazebo-yarp-plugins/pull/247 

> Btw, I am still looking for storing all these thrift files somewhere else, in a centralized place in order to avoid to keep vendoring these files in other projects.

I agree.

diegoferigo:
Ack. I'll keep in mind to update the wb-toolbox thrift in any case :+1:

PR #352, opened 2018-05-06
Add handling of remote variable 'kinematic_mj' in ControlBoard device driver

nunoguedelha:
This change allows to extract the joint coupling matrices from the `gazebo-yarp-plugin` configuration files (refer to https://github.com/robotology-playground/icub-model-generator/pull/87), populate the remote debug variable `kinematic_mj` and expose it through the interface `yarp::dev::IRemoteVariables`.


traversaro:
I think we can safely close this for the time being, as the COUPLING parameters were reverted in https://github.com/robotology/icub-model-generator/pull/88 . We can reopen when someone will be working on this, thanks!

PR #353, opened 2018-05-22
getCurrent streaming torque

lrapetti:
Accordingly to the [ControlBoardDriverCurrent](https://github.com/robotology/gazebo-yarp-plugins/blob/d8772a8f14428787a55080dae73eb55bf374746c/plugins/controlboard/src/ControlBoardDriverCurrent.cpp), as long as the current interfaces are not implemented in Gazebo, they can be used in simulation streaming the torque.

### DIFF
--- a/plugins/controlboard/src/ControlBoardDriverOthers.cpp
+++ b/plugins/controlboard/src/ControlBoardDriverOthers.cpp
@@ -112,7 +112,7 @@ bool GazeboYarpControlBoardDriver::getCurrents(double *vals) //NOT IMPLEMENTED
 {
     if (!vals) return false;
     for (size_t j = 0; j < m_numberOfJoints; ++j) {
-        vals[j] = m_torques[j];
+        this->getCurrent(j,&vals[j]);
     }
     return true;
 }


traversaro:
Thanks @lrapetti . Given that I guess that the `IAmplifierInterface` is practically abandoned (see https://github.com/robotology/yarp/issues/1694), I think we can take the chance to cleanup all the methods related to current control. 

We can: 
*  Move all the methods defined in the `ICurrentControl` interface in the `ControlBoardDriverCurrent.cpp`
* Make sure that  all the methods defined in the `ICurrentControl` interface are consistent, i.e. if we agree that the behavior of the simulated current control is to be exactly equivalent to the joint torque control, this should be the case for all methods. 
* Cleanup attributes that are not used any more after this changes, such as `m_amp`.

traversaro:
Thanks!

lrapetti:
I have modified the _gazeboYARPplugin_ on my fork https://github.com/lrapetti/gazebo-yarp-plugins/commit/110c68e3ba97227230bc2bec25b47448105b9c3f implementing the intefaces in `ControlBoardDriverCurrent.cpp` and removing the `IAmplifierInterface`.

Now the `ICurrentControl` interfaces stream the torque accordingly to a constant `m_kPWM`:
$ \tau = k_{PWM} \cdot i $

Together with the `getCurrentRange`, I have also implemented the `getTorqueRange` interface that was missing. [Now](https://github.com/lrapetti/gazebo-yarp-plugins/blob/master/plugins/controlboard/src/ControlBoardDriverTorqueControl.cpp#L99) the `getTorqueRange` is implemented giving the value of `m_maxTorques` (that was already defined in the `ControlBoardDriver.h`, but it was never used).

Further considerations have to be done on `m_maxTorques` since it is initialized at [2000 N meter](https://github.com/lrapetti/gazebo-yarp-plugins/blob/master/plugins/controlboard/src/ControlBoardDriver.cpp#L66), while for example in `yarpmotorgui.cpp`a value of [5.0](https://github.com/robotology/yarp/blob/master/src/yarpmotorgui/partitem.cpp#L201) is hard-coded as torque max value.

traversaro:
Ok, can you open a PR from your branch at https://github.com/lrapetti/gazebo-yarp-plugins/commit/110c68e3ba97227230bc2bec25b47448105b9c3f ? You can check the guide here for understanding the flow: https://guides.github.com/introduction/flow/ .

traversaro:
For more info on git, check https://github.com/vvv-school/vvv18/wiki/School-material#git-basics .

lrapetti:
This PR has been replaced by https://github.com/robotology/gazebo-yarp-plugins/pull/355.

PR #364, opened 2018-06-01
Feature/base pose velocity

prashanthr05:
For a desried link, provides the pose, velocity and acceleration of the link in the world frame of reference.

### DIFF
--- a/plugins/baseposevelocity/src/BasePoseVelocityDriver.cpp
+++ b/plugins/baseposevelocity/src/BasePoseVelocityDriver.cpp
@@ -12,7 +12,7 @@
 
 using namespace yarp::dev;
 
-GazeboYarpBasePoseVelocityDriver::GazeboYarpBasePoseVelocityDriver()
+GazeboYarpBasePoseVelocityDriver::GazeboYarpBasePoseVelocityDriver() : m_robot(0), m_baseLink(0)
 {
 
 }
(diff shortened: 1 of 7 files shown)


plugins/baseposevelocity/src/BasePoseVelocityDriver.cpp
```diff
@@ -0,0 +1,164 @@
...
+    return true;
+}
+
+void GazeboYarpBasePoseVelocityDriver::onUpdate(const gazebo::common::UpdateInfo& _info)
+{
+    ignition::math::Vector3d m_worldBasePosition = m_baseLink->WorldPose().Pos();
+    ignition::math::Quaterniond m_worldBaseOrientation = m_baseLink->WorldPose().Rot();
+    ignition::math::Vector3d m_worldBaseLinVel = m_baseLink->WorldLinearVel();
+    ignition::math::Vector3d m_worldBaseAngVel = m_baseLink->WorldAngularVel();
+    ignition::math::Vector3d m_worldBaseLinAcc = m_baseLink->WorldLinearAccel();
```
traversaro: Not super relevant, but please be aware of https://bitbucket.org/osrf/gazebo/issues/893/imusensor-should-get-acceleration-directly#comment-19435791 .

prashanthr05:
Is support still provided for gazebo7 ? Should I modify the source code to support gazebo7 ?

plugins/baseposevelocity/src/BasePoseVelocityDriver.cpp
traversaro: The first step should be open a proper issue in Gazebo, but I think it is ok not do anything, I just want you to be aware of this. Once we have a proper Gazebo issue, we can link it in the documentation.

traversaro:
> Is support still provided for gazebo7 ? Should I modify the source code to support gazebo7 ? 

Yep, Gazebo7 is the version used by ROS Kinetic and Lunar so it would be ideal to keep compatibility with it.

traversaro:
Minor and not blocking comment: I am a bit confused by the name: `baseposevelocity`, as the plugin actually also streams the acceleration. Someone has a better idea for a name @gabrielenava @diegoferigo @DanielePucci ?

prashanthr05:
Alright! I am updating the source to be gazebo7 compatible as well. Should we call it `basestate`

gabrielenava:
`baseState` seems ok to me

traversaro:
I agree (even if the acceleration is not a **state** if we consider the torques as input to the system, but I think it a legitimate poetic license).

diegoferigo:
Brainstorming: `baseInertialInfo`? Just putting here alternatives, `baseState` LGTM.

gabrielenava:
or: `baseKinematics`

traversaro:
I still prefer `basestate` to be honest.

prashanthr05:
if everyone agrees with `basestate`, I will rename, make the corresponding changes and push a commit.

DanielePucci:
Mmmmhhh I have to contain my CTI (i.e. Control Taliban Instinct) to confirm `basestate` (as @traversaro said, we shall perform system dynamical extension to consider acceleration in the system state)

But it's ok 😄 Indeed, extending the system state is what we are currently doing with @ahmadgazar, so in the next future I think it is also definitely ok

prashanthr05:
@traversaro  Sorry, did not have my computer configured for Gazebo7, used Travis as the scapegoat! Let me know, if I have to squash and rebase the commits.

traversaro:
I can squash them while merging, no problem. 

I will be merging to `devel` that will soon be release as part of the YARP 3.0 release, ok ?

PR #368, opened 2018-06-07
Fix setRefCurrent methods

lrapetti:
The `setRefCurrent` method is returning `false` when the refCurrent is zero, and the `setRefCurrents(const int n_joint, const int *joints, const double *t)` method is not using the constant `m_kPWM` converting torque to current.


traversaro:
Thanks!

PR #371, opened 2018-07-03
Improved WorldInterface::getPose(..)

valegagge:
I improved getPose in order to get the pose of an object refer to a given frame.

```
Now the signature of getPose function is:
 /**
    * Get object pose.
    * @param id string that identifies object in gazebo (returned after creation)
    * @param frame_name (optional) is specified, the pose will be relative to the specified fully scoped frame (e.g. MODEL_ID::FRAME_ID). Otherwise, world it will be used.
    * @return returns value of the pose in the world reference frame
    */
    Pose getPose(1:string id, 2: string frame_name="");
```
Moreover I fixed a small bug in makeSphere: now the objectname parameter is used to label the created sphere. see https://github.com/robotology/gazebo-yarp-plugins/commit/de34ac26bcbd8565d1d55875d7d7c2ec34b93302

This branch should solve issue https://github.com/hsp-iit/project-mobile-manipulation/issues/10

### DIFF
--- a/plugins/worldinterface/src/worldproxy.cpp
+++ b/plugins/worldinterface/src/worldproxy.cpp
@@ -892,7 +892,7 @@ bool WorldProxy::detach(const std::string& id)
 }
 
 
-GazeboYarpPlugins::Pose WorldProxy::getPose(const std::string& id)
+GazeboYarpPlugins::Pose WorldProxy::getPose(const std::string& id, const std::string& frame_name)
 {
   GazeboYarpPlugins::Pose ret;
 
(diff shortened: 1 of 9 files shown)


plugins/worldinterface/src/worldproxy.cpp
```diff
@@ -907,23 +914,47 @@ GazeboYarpPlugins::Pose WorldProxy::getPose(const std::string& id)
...
+        #if GAZEBO_MAJOR_VERSION >= 8
+        relative_link_pose = relative_link->WorldPose();
+        #else
+        relative_link_pose = relative_link->GetWorldPose();
+        #endif
+    }
+
+    #if GAZEBO_MAJOR_VERSION >= 8
+    YarpWorldPose p=model->WorldPose();
+    YarpWorldPose final_pose (p.Pos()[0], p.Pos()[1], p.Pos()[2], p.Rot().Roll(), p.Rot().Pitch(), Rot().Yaw());
```
miccol: `Rot().Yaw()`  should be replaced with  `p.Rot().Yaw()`, right?

miccol:
Travis does not compile due to the`ALLOW_IDL_GENERATION` set to `OFF`.
I tried on my PC (Ubuntu 16.04 and Gazebo 8.6) and, besides my comment above, works fine (checked both relative position and orientation).

Thank you!

traversaro:
Autogenerated files generated by the `yarp_idl_to_dir` macro and generated in the source directory should be committed to the repository, so people can use the software without changing the `ALLOW_IDL_GENERATION` variable, see https://github.com/robotology/yarp/issues/368 .

plugins/worldinterface/src/worldproxy.cpp
valegagge: Done! see 56ef8787ebd357a395ecc822e90acf8c20dd99b9
Sorry for the delay!

valegagge:
About the autogenerated code, I think I commited all already.
Could you tell me which changes I forgot? @miccol 
thank you!

miccol:
@valegagge Is right, the autogenerated code is there. I am not sure why I could compile only with the IDL generation ON.

Travis is failing due to the YARP version (Installed: 3.0.100, required 3.0.101).
I tried on my local PC and the PR works fine!
Thank you

MC

traversaro:
Ok, if no one objects I will merge this evening!

Issue #373, opened 2018-07-13
Gazebo-yarp-plugins devel non opening ports


lrapetti:
I have compiled the `gazebo-yarp-plugins` in `devel`, but when I open a model I don't see all the ports opened:

```
registration name /clock ip 192.168.1.5 port 10002 type tcp
registration name /clock/rpc ip 192.168.1.5 port 10003 type tcp
registration name /icub02 ip 192.168.1.5 port 10000 type tcp
registration name /icubSim/applyExternalWrench/rpc:i ip 192.168.1.5 port 10017 type tcp
registration name /icubSim/inertial ip 192.168.1.5 port 10004 type tcp
registration name /icubSim/left_arm/analog:o ip 192.168.1.5 port 10014 type tcp
registration name /icubSim/left_arm/analog:o/rpc:i ip 192.168.1.5 port 10013 type tcp
registration name /icubSim/left_foot/analog:o ip 192.168.1.5 port 10008 type tcp
registration name /icubSim/left_foot/analog:o/rpc:i ip 192.168.1.5 port 10007 type tcp
registration name /icubSim/left_leg/analog:o ip 192.168.1.5 port 10006 type tcp
registration name /icubSim/left_leg/analog:o/rpc:i ip 192.168.1.5 port 10005 type tcp
registration name /icubSim/right_arm/analog:o ip 192.168.1.5 port 10016 type tcp
registration name /icubSim/right_arm/analog:o/rpc:i ip 192.168.1.5 port 10015 type tcp
registration name /icubSim/right_foot/analog:o ip 192.168.1.5 port 10012 type tcp
registration name /icubSim/right_foot/analog:o/rpc:i ip 192.168.1.5 port 10011 type tcp
registration name /icubSim/right_leg/analog:o ip 192.168.1.5 port 10010 type tcp
registration name /icubSim/right_leg/analog:o/rpc:i ip 192.168.1.5 port 10009 type tcp
registration name fallback ip 224.2.1.1 port 10000 type mcast
*** end of message
```

lrapetti:
Solved by https://github.com/robotology/gazebo-yarp-plugins/pull/374

Issue #372, opened 2018-07-13
Linking of GazeboYARPPlugins devel in robotology-superbuild fails.


nunoguedelha:
### Issue description

This issue was introduced by commit 8bf4e24ada037d99d1fa50c67fc9239d012675bd. When building and linking the component within the superbuild we get the following error:
```
Ld /Users/nunoguedelha/dev/robotology-superbuild/build/robotology/GazeboYARPPlugins/plugins/controlboard/Release/libgazebo_yarp_controlboard.dylib normal x86_64
    cd /Users/nunoguedelha/dev/robotology-superbuild/robotology/GazeboYARPPlugins
    export MACOSX_DEPLOYMENT_TARGET=10.12
...
...
ld: warning: directory not found for option '-L/usr/local/Cellar/gazebo8/8.6.0/lib/Release'
ld: warning: directory not found for option '-L/usr/local/Cellar/gazebo8/8.6.0/lib/gazebo-8/plugins/Release'
ld: warning: directory not found for option '-L/usr/local/Cellar/bullet/2.87/lib/Release'
ld: warning: directory not found for option '-L/usr/local/lib/Release'
ld: warning: directory not found for option '-L/usr/local/Cellar/ignition-transport3/3.1.0_1/lib/Release'
ld: warning: directory not found for option '-L/usr/local/Cellar/zeromq/4.2.5/lib/Release'
ld: warning: directory not found for option '-L/usr/local/Cellar/ossp-uuid/1.6.2_2/lib/Release'
Undefined symbols for architecture x86_64:
  "yarp::dev::GazeboYarpControlBoardDriver::getMotorTorqueParams(int, yarp::dev::MotorTorqueParameters*)", referenced from:
      vtable for yarp::dev::GazeboYarpControlBoardDriver in ControlBoardDriver.o
  "yarp::dev::GazeboYarpControlBoardDriver::setMotorTorqueParams(int, yarp::dev::MotorTorqueParameters)", referenced from:
      vtable for yarp::dev::GazeboYarpControlBoardDriver in ControlBoardDriver.o
  "non-virtual thunk to yarp::dev::GazeboYarpControlBoardDriver::getMotorTorqueParams(int, yarp::dev::MotorTorqueParameters*)", referenced from:
      vtable for yarp::dev::GazeboYarpControlBoardDriver in ControlBoardDriver.o
[...]

PR #374, opened 2018-07-17
Add setMotorTorqueParams and getMotorTorqueParams methods to the controlboard class

traversaro:
It was erroneously removed in  https://github.com/robotology/gazebo-yarp-plugins/commit/8bf4e24ada037d99d1fa50c67fc9239d012675bd#diff-732527c53bcfabfbb15e5399654ff8b6L123 .

Fix https://github.com/robotology/gazebo-yarp-plugins/issues/372 . 
Fix https://github.com/robotology/gazebo-yarp-plugins/issues/373 .

### DIFF
--- a/plugins/controlboard/src/ControlBoardDriverTorqueControl.cpp
+++ b/plugins/controlboard/src/ControlBoardDriverTorqueControl.cpp
@@ -109,6 +109,16 @@ bool GazeboYarpControlBoardDriver::getTorqueRanges(double *min, double *max)
     return true;
 }
 
+bool GazeboYarpControlBoardDriver::getMotorTorqueParams(int ,  yarp::dev::MotorTorqueParameters *)
+{
+    return false;
+}
+
+bool GazeboYarpControlBoardDriver::setMotorTorqueParams(int , const yarp::dev::MotorTorqueParameters )
+{
+    return false;
+}
+
 bool GazeboYarpControlBoardDriver::checkIfTorqueIsValid(const double* torques) const
 {
     if (!torques)


traversaro:
The Travis was failing due to the outdated docker build, that at the moment it is not working due to the YARP situation ( https://github.com/robotology/QA/issues/300 ) .

randaz81:
Just for curiosity: was this error a MAC-related error?
Because I cannot reproduce it on my gcc.

nunoguedelha:
> For me, using Ubuntu, it was compiling but it was not opening correctly the controlBoard when opening the model #373

as per [@traversaro 's comment](https://github.com/robotology/gazebo-yarp-plugins/issues/372#issuecomment-405509500) it was related to an undefined symbol, so probably the build link phase went wrong.

Maybe on MacOs the compiler is less permissive so raises the issue at link time.

traversaro:
> Maybe on MacOs the compiler is less permissive so raises the issue at link time.

Exactly (it is actually related to the difference between [ELF](https://en.wikipedia.org/wiki/Executable_and_Linkable_Format) and  [Mach-O](https://en.wikipedia.org/wiki/Mach-O) executable file formats. On ELF some symbols are not resolved until a shared library is linked to an executable is actually created, or a shared library is loaded using `dlopen`. 

You can actually change the linker setting on Linux to make sure that is behaves like macOs , as done by @diegoferigo in https://github.com/robotology/wb-toolbox/commit/69ce602978b4e1494fe9460780e8e9fef1e8878a . See https://github.com/robotology/wb-toolbox/issues/112#issuecomment-387433942 .

jeljaik:
Thanks, I was having the same issue on my side.

PR #375, opened 2018-07-17
Add setControlMode and setControlModes methods dummy definitions

nunoguedelha:
Same fix as https://github.com/robotology/gazebo-yarp-plugins/pull/374, for the same reasons. These method definitions were erroneously removed in  https://github.com/robotology/gazebo-yarp-plugins/commit/8bf4e24ada037d99d1fa50c67fc9239d012675bd#diff-732527c53bcfabfbb15e5399654ff8b6L123 .

### DIFF
--- a/plugins/fakecontrolboard/src/FakeControlBoardDriverControlMode.cpp
+++ b/plugins/fakecontrolboard/src/FakeControlBoardDriverControlMode.cpp
@@ -41,7 +41,16 @@ bool GazeboYarpFakeControlBoardDriver::getControlMode(int j, int *mode)
     return true;
 }
 
-bool GazeboYarpFakeControlBoardDriver::setControlModes(const int n_joint, const int *joints, int *modes) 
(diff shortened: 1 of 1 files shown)


traversaro:
Thanks, Travis is failing due to the outdated docker image, that is not easy to fix due to https://github.com/robotology/QA/issues/300 .

nunoguedelha:
yes, I just saw your comment on the other PR. Thx

randaz81:
Just for curiosity: was this error a MAC-related error?
Because I cannot reproduce it on my gcc.

nunoguedelha:
I would say it isn't MACOS specific since the issue appears under another form in Linux (https://github.com/robotology/gazebo-yarp-plugins/pull/373).

traversaro:
> Just for curiosity: was this error a MAC-related error?
Because I cannot reproduce it on my gcc.

Not really, on GCC/Ubuntu the problem is just delayed on when you launched a model in Gazebo that loaded this plugin, see https://github.com/robotology/gazebo-yarp-plugins/pull/374#issuecomment-405585759 .

PR #378, opened 2018-07-26
Fix laser plugin

valegagge:
When I tried to read laser values in Gazebo I couldn't read nothing because of the following error:
`[ERROR]Cannot find "gazebo_lasersensor" plugin (not built in, and no .ini file found for it)
[ERROR]Check that YARP_DATA_DIRS leads to at least one directory with plugins/gazebo_lasersensor.ini or share/yarp/plugins/gazebo_lasersensor.ini in it
yarpdev: ***ERROR*** could not find device <gazebo_lasersensor>
[ERROR]RangeFinder2DWrapper: failed to open subdevice.. check params 
yarpdev: ***ERROR*** driver <Rangefinder2DWrapper> was found but could not open
[ERROR]GazeboYarpLaserSensor Plugin failed: error in opening yarp driver wrapper 
`

Therefore I try to fix this issue con this PR.
Now Laser works fine on my setup. 
@barbalberto could you check if this fix is correct?
Thanks!

### DIFF
--- a/plugins/lasersensor/src/LaserSensor.cc
+++ b/plugins/lasersensor/src/LaserSensor.cc
@@ -54,7 +54,7 @@ void GazeboYarpLaserSensor::Load(sensors::SensorPtr _sensor, sdf::ElementPtr _sd
 
     // Add my gazebo device driver to the factory.
     ::yarp::dev::Drivers::factory().add(new ::yarp::dev::DriverCreatorOf< ::yarp::dev::GazeboYarpLaserSensorDriver>
-                                      ("gazebo_laserSensor", "Rangefinder2DWrapper", "GazeboYarpLaserSensorDriver"));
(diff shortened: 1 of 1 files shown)


valegagge:
Ok. Now I can't check it, but I'll do asap.😊

valegagge:
@barbalberto .
I did the changes you suggested and now the laser work.

Therefore how do we procede? 
Do I commit the changes in .ini file and removed the last commit?
When we talked f2f, you proposed also to uncomment the line
https://github.com/robotology/gazebo-yarp-plugins/blob/17230ad6141a16e9cd5396703a37ddad2b346ea2/plugins/lasersensor/src/LaserSensor.cc#L93 and use the correct name of device like alternative to ini file changes.

valegagge:
@barbalberto ping....
any news about this? I need to move on

barbalberto:
Ok, after chatting also with @randaz81, we agreed to fix the config file.
Later on we will re-check the plugins code, to see if there is a way to make things easier for the user and avoid such mistakes/confusion.

I'll close this PR.

PR #390, opened 2018-10-05
Fix typo prohibiting build with gazebo<8

Tobias-Fischer:


traversaro:
Unfortunately our CI is not running aganinst Gazebo7, as there are now corresponding docker images in https://hub.docker.com/r/robotology/robotology-tdd/ .

Issue #342, opened 2018-01-30
Port naming conflict while spawning multiple robots through a gazebo world file 


yeshasvitirupachuri:
As mentioned in this #93 while spawning multiple icub robots into gazebo `gazeboYarpPluginsRobotName icubSim` inside `gazebo_icub_robotname.ini` is commented out. 

All the ports open correctly while spawning(dragging and dropping) two base icub models inside gazebo

```
registration name /iCub/head/command:i ip 10.255.54.112 port 10201 type tcp
registration name /iCub/head/rpc:i ip 10.255.54.112 port 10200 type tcp
registration name /iCub/head/state:o ip 10.255.54.112 port 10202 type tcp
registration name /iCub/head/stateExt:o ip 10.255.54.112 port 10203 type tcp
registration name /iCub/inertial ip 10.255.54.112 port 10004 type tcp
registration name /iCub/left_arm/analog:o ip 10.255.54.112 port 10008 type tcp
registration name /iCub/left_arm/analog:o/rpc:i ip 10.255.54.112 port 10007 type tcp
registration name /iCub/left_arm/command:i ip 10.255.54.112 port 10205 type tcp
registration name /iCub/left_arm/rpc:i ip 10.255.54.112 port 10204 type tcp
registration name /iCub/left_arm/state:o ip 10.255.54.112 port 10206 type tcp
registration name /iCub/left_arm/stateExt:o ip 10.255.54.112 port 10207 type tcp
registration name /iCub/left_foot/analog:o ip 10.255.54.112 port 10191 type tcp
registration name /iCub/left_foot/analog:o/rpc:i ip 10.255.54.112 port 10190 type tcp
registration name /iCub/left_leg/analog:o ip 10.255.54.112 port 10189 type tcp
registration name /iCub/left_leg/analog:o/rpc:i ip 10.255.54.112 port 10188 type tcp
registration name /iCub/left_leg/command:i ip 10.255.54.112 port 10213 type tcp
registration name /iCub/left_leg/rpc:i ip 10.255.54.112 port 10212 type tcp
registration name /iCub/left_leg/state:o ip 10.255.54.112 port 10214 type tcp
registration name /iCub/left_leg/stateExt:o ip 10.255.54.112 port 10215 type tcp
registration name /iCub/right_arm/analog:o ip 10.255.54.112 port 10010 type tcp
registration name /iCub/right_arm/analog:o/rpc:i ip 10.255.54.112 port 10009 type tcp
[...]

Issue #142, opened 2014-10-14
Add option for simple configuration for the plugins


traversaro:
For sufficiently simple use the plugin should be configurable just from the sdf, without the need of a `.ini` file.

traversaro:
Just realized how we can solve this issue in two lines of code. We normally load yarp-specific configuration files using the following syntax:
~~~
<yarpConfigurationFile>model://icub/conf/FT/gazebo_icub_left_leg_ft.ini</yarpConfigurationFile>
~~~
where the content of the tag is the path that is passed to the `yarp::os::Property::fromConfigFile` method.

We can add as an alternative option this tag:
~~~
<yarpConfigurationString>"(period 0.01)"</yarpConfigurationString> 
~~~
that passes its content directly to yarp::os::Property::fromString`.

lrapetti:
@traversaro 
Given that I may need this feature (see #393) I could take over this task, but I would like to discuss first what would be the relationship between  `yarpConfigurationFile` and `yarpConfigurationString`:
  -  The two element can coexist and the properties is the result of combining the properties from the file and from the string (what to do in case a property is defined in both, which one would have the priority)
  - The two element can coexist, but the properties are read only from one of the two. 
  - Fail if the two properties coexist

(Probably this should be decided having in view the possibly overriding a property as in #392)

traversaro:
For a single plugin, I would say that the best option is that the two properties cannot coexist, i.e. an error is printed and the plugins does not start. When overriding, by idea was something like: Only  `yarpConfigurationString` is permitted during overriding, and the loading is structured in this way (similar to https://github.com/robotology/yarp/blob/19cdb60478371d4cba8b9889c5f582d6bb5e6fc7/src/libYARP_OS/src/ResourceFinder.cpp#L204):  
first `yarpConfigurationFile` is passed to `yarp::os::Property::fromConfigFile`, and then the overriding `yarpConfigurationString` using the `fromString(.. , false)` to just override the contained parameters (or perhaps even better using the `fromCommand(..., false)` that is what is used for parameter overriding in RFModules, so it is already extensively tested, and I guess the syntax `--param1 1.0` is more intuitive then `(param1 1.0)`. 

I do not how much this is compatible with the `libgazebo_yarp_configurationoverride.so` plugin idea outline in https://github.com/robotology/gazebo-yarp-plugins/issues/392 , while it could be doable modifying the plugin source code.  

[...]

PR #393, opened 2018-10-25
Refactoring of linkattacher plugin

lrapetti:
As it was originally implemented, the `linkattacher` device was though to attach the link of a robot (`robot_model`) to that of an object (`object_model`).

In order to have a general device (to be used for example in order to simulate contacts in robot-robot interaction), it was not only necessary to change the name of the variables to more proper names (`parent_model` and `child_model`) but it was also necessary to be able to detect the link of both the models from scoped names (to understand why see https://github.com/robotology/gazebo-yarp-plugins/issues/342 ) instead of using [`GetLink()`](https://github.com/robotology/gazebo-yarp-plugins/compare/devel...lrapetti:linkattacher_refactoring?expand=1#diff-7e1089c30889bd5585eeb97aae62e840L38).

In addiction, I have used the [hasEnding()](https://github.com/robotology/gazebo-yarp-plugins/blob/master/libraries/common/include/GazeboYarpPlugins/common.h#L52) in place of the previous implementation of the search of the link from unscoped name, in this way it is possible to use both the scoped or unscoped name as input (this can be helpful to distinguish between links with the same unscoped name).

#### Further Implementation
Given that the search for the link from `model_name` and `link_name` is repeated along the code, and could be useful also for other plugins, I was thinking that it would be better to have this function defined only once in a shared place ([`common.h`](https://github.com/robotology/gazebo-yarp-plugins/blob/master/libraries/common/include/GazeboYarpPlugins/common.h)??) as something like:
```
gazebo::physics::LinkPtr GetLinkFromName(const string& model_name, const string& link_name)
```

### DIFF
--- a/plugins/linkattacher/src/linkattacherserverimpl.cpp
+++ b/plugins/linkattacher/src/linkattacherserverimpl.cpp
@@ -16,172 +16,179 @@ LinkAttacherServerImpl::LinkAttacherServerImpl()
 LinkAttacherServerImpl::~LinkAttacherServerImpl()
 {}
 
-bool LinkAttacherServerImpl::attachUnscoped(const string& model_name, const std::string& model_link_name, const std::string& robot_name, const string& robot_link_name)
+bool LinkAttacherServerImpl::attachUnscoped(const string& parent_model_name, const string& parent_model_link_name, const string& child_model_name, const string& child_model_link_name)
 {
-
     #if GAZEBO_MAJOR_VERSION >= 8
-    gazebo::physics::ModelPtr object_model = _world->ModelByName(model_name);
+    gazebo::physics::ModelPtr parent_model = _world->ModelByName(parent_model_name);
     #else
-    gazebo::physics::ModelPtr object_model = _world->GetModel(model_name);
+    gazebo::physics::ModelPtr parent_model = _world->GetModel(parent_model_name);
     #endif
 
-    if(!object_model)
+    if(!parent_model)
     {
-        yError() << LogPrefix << "Attach: " << model_name << " does not exist in gazebo";
+        yError() << LogPrefix << "Attach: " << parent_model_name << " model does not exist in gazebo";
         return false;
     }
     else
     {
-      yInfo() << LogPrefix << "Attach: " << object_model->GetName() << " found";
-    }
-
-    gazebo::physics::LinkPtr object_link = object_model->GetLink(model_link_name);
-    if(!object_link)
-    {
-        yError() << LogPrefix << "Attach: " << model_link_name << " is not found";
-        return false;
-    }
-    else
-    {
-      yInfo() << LogPrefix << "Attach: " << object_link->GetName() << " found";
-    }
-
-    #if GAZEBO_MAJOR_VERSION >= 8
-    gazebo::physics::ModelPtr robot_model = _world->ModelByName(robot_name);
-    #else
-    gazebo::physics::ModelPtr robot_model = _world->GetModel(robot_name);
-    #endif
-
-    if(!robot_model)
-    {
-        yError() << LogPrefix << "Attach: " << robot_name << " does not exist in gazebo";
-        return false;
-    }
-    else
-    {
-      yInfo() << LogPrefix << "Attach: " << robot_model->GetName() << " found";
+      yInfo() << LogPrefix << "Attach: " << parent_model->GetName() << " model found";
     }
 
     //Get the exact link with only link name instead of full_scoped_link_name
-    gazebo::physics::Link_V robot_model_links = robot_model->GetLinks();
-    for(int i=0; i < robot_model_links.size(); i++)
+    gazebo::physics::Link_V parent_model_links = parent_model->GetLinks();
+    gazebo::physics::LinkPtr parent_model_link;
+    for(int i=0; i < parent_model_links.size(); i++)
     {
-        //E.g iCub::l_hand::l_hand_base_link
-
-        std::string candidate_robot_link_name = robot_model_links[i]->GetScopedName();
-
-        std::size_t lastcolon = candidate_robot_link_name.rfind(":");
-        std::string unscoped_robot_link_name = candidate_robot_link_name.substr(lastcolon+1,std::string::npos);
-
-        if(unscoped_robot_link_name == robot_link_name)
+        std::string candidate_parent_model_link_name = parent_model_links[i]->GetScopedName();
+        
+        // hasEnding compare ending of the condidate model link name to the given link name, in order to be able to use unscoped names
+        if(GazeboYarpPlugins::hasEnding(candidate_parent_model_link_name, parent_model_link_name))
         {
-            yInfo() << LogPrefix << "Attach: Full scoped Candidate robot link name: " << candidate_robot_link_name << " found";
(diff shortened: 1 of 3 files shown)


plugins/linkattacher/src/linkattacherserverimpl.cpp
```diff
@@ -16,147 +16,155 @@ LinkAttacherServerImpl::LinkAttacherServerImpl()
...
+    joint->Attach(parent_model_link,child_model_link);
 
     return true;
 }
 
 bool LinkAttacherServerImpl::detachUnscoped(const string& model_name, const std::string& model_link_name)
 {
-    #if GAZEBO_MAJOR_VERSION >= 8
-    gazebo::physics::ModelPtr object_model = _world->ModelByName(model_name);
+	#if GAZEBO_MAJOR_VERSION >= 8
```
traversaro: Is this  a tab?

traversaro:
I do not think I can handle this review. @Yeshasvitvs can you review it? Once you approve, I will be happy to merge.

@lrapetti not sure if there is any need, but perhaps you may want to edit https://github.com/robotology/gazebo-yarp-plugins/blob/cac305ccae3e38640eaa6a966bc6cfa58722fd33/plugins/linkattacher/README.md to reflect the changes in the plugin.

plugins/linkattacher/src/linkattacherserverimpl.cpp
```diff
@@ -16,147 +16,155 @@ LinkAttacherServerImpl::LinkAttacherServerImpl()
...
+    	return false;
+    }
 
-            #if GAZEBO_MAJOR_VERSION >= 8
-            joint = _world->Physics()->CreateJoint("fixed",object_model);
-            #else
-            joint = _world->GetPhysicsEngine()->CreateJoint("fixed",object_model);
-            #endif
+    //This is joint creation
+    gazebo::physics::JointPtr joint;
```
Yeshasvitvs: I think it i better to use the variable name to `fixedJoint` or `magnetJoint` to be more reflective of its use.

plugins/linkattacher/src/linkattacherserverimpl.cpp
```diff
@@ -16,147 +16,155 @@ LinkAttacherServerImpl::LinkAttacherServerImpl()
...
+    #else
+    joint = _world->GetPhysicsEngine()->CreateJoint("fixed",parent_model);
+    #endif
 
-            std::string joint_name = model_link_name + "_magnet_joint";
-            joint->SetName(joint_name);
-            yInfo() << LogPrefix << "Attach: Magnet joint: " << joint->GetName() << " created";
+    if(!joint)
+    {
+        yError() << LogPrefix << "Attach: Unable to create joint";
```
Yeshasvitvs: This should be consistent with the joint name change above

plugins/linkattacher/src/linkattacherserverimpl.cpp
```diff
@@ -16,147 +16,155 @@ LinkAttacherServerImpl::LinkAttacherServerImpl()
...
+        return false;
+    }
 
-            joint->SetModel(object_model);
+    std::string joint_name = parent_model_link_name + "_magnet_joint";
```
Yeshasvitvs: The `joint_name` variable name should be consistent with the joint name change above

plugins/linkattacher/src/linkattacherserverimpl.cpp
```diff
@@ -16,147 +16,155 @@ LinkAttacherServerImpl::LinkAttacherServerImpl()
...
+    #if GAZEBO_MAJOR_VERSION >= 8
+    joint->Load(parent_model_link,child_model_link,ignition::math::Pose3d());
+    #else
+    joint->Load(parent_model_link,child_model_link,gazebo::math::Pose());
+    #endif
 
-            break;
-        }
-    }
+    //Attach(prent_link,child_link)
```
Yeshasvitvs: my bad `parent` misspelled  :see_no_evil:

plugins/linkattacher/src/linkattacherserverimpl.cpp
```diff
@@ -16,147 +16,155 @@ LinkAttacherServerImpl::LinkAttacherServerImpl()
...
+    //Get the exact link with only link name instead of full_scoped_link_name
+    gazebo::physics::Link_V model_links = model->GetLinks();
+    gazebo::physics::LinkPtr model_link;
+    for(int i=0; i < model_links.size(); i++)
     {
-        yError() << LogPrefix << "Detach: " << model_link_name << " is not found";
-        return false;
+        std::string candidate_model_link_name = model_links[i]->GetScopedName();
+
+        if(GazeboYarpPlugins::hasEnding(candidate_model_link_name, model_link_name))
```
Yeshasvitvs: Please add the comment for `hasEnding()` function

Yeshasvitvs:
@traversaro I reviewed the PR and just some minor changes are needed but other than that LGTM! :+1:

lrapetti:
Updated according to @Yeshasvitvs review

lrapetti:
Can we merge this?

lrapetti:
> Can we merge this?

Actually before merging I have another propose of improvement.
Currently, the device is loading the port name from a configuration file as described [here](https://github.com/robotology/gazebo-yarp-plugins/tree/master/plugins/linkattacher#usage), and this means creating an `.ini` for each model-name.

Is there any reason for having an `.ini` just for storing this information?
Otherwise, it could be useful to also have the possibility of defining the port name also directly in the sdf without passing trough another file:
```xml
<model name="iCub">
    <include>
        <uri>model://icub_standup</uri>
    </include>
    <plugin name='link_attacher' filename='libgazebo_yarp_linkattacher.so'>
      <rpcName>/iCub/linkattacher/rpc:i</rpcName>
    </plugin>
</model>
```

traversaro:
Related issue: https://github.com/robotology/gazebo-yarp-plugins/issues/142 . Maintaining both a SDF-based configuration and a yarp .ini-conifguration may be a mantainance nigthmare,  so I think https://github.com/robotology/gazebo-yarp-plugins/issues/142 may be interesting. 

> Can we merge this?

I am  waiting for  @Yeshasvitvs approval (or any other reviewer, if @Yeshasvitvs is currently blocked).

Yeshasvitvs:
@traversaro the changes look good to me, you can proceed to merge.

lrapetti:
> Related issue: #142 . Maintaining both a SDF-based configuration and a yarp .ini-conifguration may be a maintenance nightmare, so I think #142 may be interesting.

@traversaro 
Nice, I could start by adding the `yarpConfigurationString` parameter to the `linkattacher` and then eventually it could be add to all the plugins using `yarpConfigurationFile`

lrapetti:
@traversaro 
We got the approval and we may be able to merge it,

The following enhancement is not specific of `linkattacher` and will be discussed directly in #142
> > Related issue: #142 . Maintaining both a SDF-based configuration and a yarp .ini-conifguration may be a maintenance nightmare, so I think #142 may be interesting.
> 
> @traversaro
> Nice, I could start by adding the `yarpConfigurationString` parameter to the `linkattacher` and then eventually it could be add to all the plugins using `yarpConfigurationFile`

PR #394, opened 2018-11-09
Add use of loadConfigModelPlugin()

lrapetti:
As discussed in https://github.com/robotology/gazebo-yarp-plugins/issues/142#issuecomment-437087190, after introducing [GazeboYarpPlugins::loadConfigModelPlugin()](https://github.com/robotology/gazebo-yarp-plugins/blob/master/libraries/singleton/src/ConfHelpers.cc#L59) with https://github.com/robotology/gazebo-yarp-plugins/commit/1528523e1b1d193dfb5aabd542b80203f57914a1, the plugins have never been cleaned up and this function was almost unused (it was used just in `jointsensors` and `inertialmtbPart`, while it was done consistently with `loadConfigSensorPlugin()` in https://github.com/robotology/gazebo-yarp-plugins/commit/1528523e1b1d193dfb5aabd542b80203f57914a1).

A spread use of `loadConfigModelPlugin()` will facilitate changes in the way parameters are loaded, so with this PR I am updating all the plugins in order to use it.
The update plugins are:
- `GazeboYarpBaseState`
- `GazeboYarpContactLoadCellArray `
- `GazeboYarpControlBoard`
- `GazeboYarpFakeControlBoard`
- `GazeboYarpLinkattacher`
- `GazeboYarpMaisSensor`
- `GazeboYarpVideoTexture`

#### Observation
- In the [`GazeboYarpControlBoard`](https://github.com/robotology/gazebo-yarp-plugins/pull/394/files#diff-350377892a04885f201132a7178fbec9L85) and [`GazeboYarpFakeControlBoard`](https://github.com/robotology/gazebo-yarp-plugins/pull/394/files#diff-41ad1594192efebe9071a061089932bbL85), a property containing the path of the `.ini` configuration file was add to the `parameters` as `"gazebo_ini_file_path"`. This operation was not replicated in any of the other plugins and it is not implemented in `loadConfigModelPlugin()`. However, the `"gazebo_ini_file_path"` property seems to be unused so removing it should not be a problem.
- In `GazeboYarpVideoTexture` the print functions were commented out, and using `loadConfigModelPlugin()` the print will be added again. Since I don't know why they were commented out, I don't know it this could be a problem. Furthermore, in the `GazeboYarpVideoTexture` the [`addGazeboEnviromentalVariablesModel `](https://github.com/robotology/gazebo-yarp-plugins/blob/master/libraries/singleton/src/ConfHelpers.cc#L67) function was previously not called and it is introduced by the `loadConfigModelPlugin()`, but this should just add some unused variable but it should not cause any problem (I have no example using this function so I was not able to test it).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,4 +19,5 @@ Mirko Ferrati <mirko.ferrati@gmail.com>
 Nicola Piga <nicolapiga@gmail.com>
 Prashanth Ramadoss <prashanth.ramadoss@iit.it>
 Silvio Traversaro <silvio.traversaro@iit.it>
-Yeshasvi Tirupachuri<yeshasvi.tirupachuri@iit.it>
+Yeshasvi Tirupachuri <yeshasvi.tirupachuri@iit.it>
+Lorenzo Rapetti <lorenzo.rapetti@iit.it>
--- a/plugins/basestate/src/BaseState.cc
+++ b/plugins/basestate/src/BaseState.cc
@@ -65,33 +65,27 @@ namespace gazebo
         ::yarp::dev::Drivers::factory().add(new ::yarp::dev::DriverCreatorOf< ::yarp::dev::GazeboYarpBaseStateDriver>
                                         ("gazebo_basestate", "analogServer", "GazeboYarpBaseState"));
         
-        yarp::os::Bottle networkDeviceProp;
-        yarp::os::Bottle deviceDriverProp;
-        
-        if (_sdf->HasElement("yarpConfigurationFile"))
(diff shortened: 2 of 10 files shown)


lrapetti:
@traversaro

lrapetti:
I did the requested changes in the style and I have corrected the typo error for `VideoTexture` in https://github.com/robotology/gazebo-yarp-plugins/pull/394/commits/bea7588f829f49809e025349682ee348c1dde7b1

traversaro:
Thanks, merging.

Issue #392, opened 2018-10-23
YARP-indipendent plugin for setting initial joint position


traversaro:
This is based on discussion that we had some time ago with @nunoguedelha . 

## Problem 
We would like to specify initial joint positions for included models, something like:
~~~xml
<include>
    <uri>model://icub</uri>
    <pose>0.0 0 0.6 0 0 3.14</pose>
    <!-- Specify joint position- -> 
     ... 
</include>
~~~

## Current situation 
There is a Gazebo issue open on this, see https://bitbucket.org/osrf/gazebo/issues/1072/allow-to-set-initial-joint-pose-in-sdf , and the `gazebo_yarp_controlboard` plugin offers a similar feature using the `initialConfiguration` option https://github.com/robotology/gazebo-yarp-plugins/blob/01ad1d1252b0827033fd57058f066cc790b893c6/plugins/controlboard/src/ControlBoardDriver.cpp#L353 . However, there is no way at the moment to overwrite this parameter in an external model. 

## Possible solution 
I think it is possible to workaround this situation by exploiting the order in which the `Load` method of a ModelPlugin](http://gazebosim.org/tutorials?tut=plugins_model 

## Open problems 
* [ ] How to specify the joint configuration in SDF (I suggest **not** to assume any predefined serialization for the joint of the robot, and instead explicitly indicate the serialization used in someway, possible mapping names to the corresponding included model using  the `hasEnding` method: https://github.com/robotology/gazebo-yarp-plugins/blob/master/libraries/common/include/GazeboYarpPlugins/common.h#L52 .
* [ ] At the moment  with `gazebo-yarp-plugins` we tipically `include` the model, we should check how thing works with "nested" models, see http://gazebosim.org/tutorials?tut=nested_model&cat=build_robot

lrapetti:
Possible idea for the conf format:
~~~
 <model name="iCub_fixed">
<include>
    <uri>model://icub</uri>
    <pose>0.0 0 0.6 0 0 3.14</pose>
    <!-- Specify joint position - -> 
</include>
    <plugin name='initial_joints' filename='libgazebo_yarp_initialjoint.so'>
         <joint_initial_position suffix="sdaddas"> 0.0</joint>
         <joint_initial_position suffix="sdaddas"></joint>
         <joint_initial_position suffix="sdaddas"> 0.0</joint>
    </plugin>
</model>
~~~

barbalberto:
[...]

PR #401, opened 2018-11-19
Add configurationOverride Plugin

lrapetti:
Accordingly to what discussed in #392, with this PR I am adding a plugin that enables to modify the value of properties of another plugin attached to the same `model`.

The plugin was initially thought to set the initial position of a model from a world or from a wrapping model, without the need of generating multiple `sdf` with the same model just to change some parameters (e.g. [`icub`](https://github.com/robotology/icub-gazebo/blob/master/icub/icub.sdf) and [`icub_standup`](https://github.com/robotology/icub-gazebo-wholebody/blob/master/models/icub_standup/icub_standup.sdf)).

The details on the use of the plugin are contained in the [`README.MD`](https://github.com/robotology/gazebo-yarp-plugins/compare/devel...lrapetti:configuration_override?expand=1#diff-326327ddda23001662d7b983d25d42de).


traversaro:
Can you bump the patch version ( https://github.com/robotology/gazebo-yarp-plugins/blob/master/CMakeLists.txt#L11 ) to 101? See https://github.com/robotology/yarp/blob/master/.github/CONTRIBUTING.md#example for the rationale.

plugins/configurationoverride/README.md
```diff
@@ -0,0 +1,27 @@
...
+    <plugin name='configuration_override' filename='libgazebo_yarp_configurationoverride'>
+      <yarpPluginConfigurationOverride plugin_name='controlboard_right_arm'> </yarpPluginConfigurationOverride>
+      <initialConfiguration>-2.0 0.52 0.0 0.785 0 0 0.698</initialConfiguration>
+    </plugin>
+    <include>
+      <uri>model://icub</uri>
+    </include>
+    </model>
+```
+then instead of loading the model `icub` with the standard right arm `<initial Configuration>`, the configuration defined in the ConfigurationOverride will be used.
```
traversaro:
```suggestion
then instead of loading the model `icub` with the standard right arm `<initialConfiguration>`, the configuration defined in the ConfigurationOverride will be used.
```

plugins/configurationoverride/src/ConfigurationOverride.cc
```diff
@@ -0,0 +1,89 @@
...
+                {
+                    // if the element to be overritten is not found in the overriden_plugin, create the element
+                    if(! overriden_plugin->HasElement(new_element->GetName()))
+                    {
+                        // AddElement
+                        sdf::ElementPtr created_element(new sdf::Element);
+                        overriden_plugin->AddElementDescription(new_element);
+                        overriden_element = overriden_plugin->GetElement(new_element->GetName());
+
+                        yInfo() << "ConfigurationOverride : " << new_element->GetName() << " has been created in " << overriden_plugin_name;
```
traversaro: I know that the existing plugin are already quite verbose in their output, but to be honest I am not a big fan of also adding this: the loading of complex models such as `iCub` are already quite verbose (to the point that if an error occurs is almost impossible to catch it) and  this will make the situation even works. @lrapetti what do you think?

lrapetti: If we imagine using the plugin to change the initial configuration for all the control boards, I agree it would be better avoid having those `yInfo()`. I can remove it.

traversaro: That is indeed how it will be used in models such as `icub_on_chair`. It can also be made optional based on some option, but the important thing for me is that it is not on by default.

lrapetti: I will just remove them, I don't think it's worth to add special cases with the print option.

plugins/configurationoverride/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+N.B.
+If a parameter is defined multiple times inside a plugin, only the first occurence will be overritten accordingly to the element retreived with [`GetElement()`](http://osrf-distributions.s3.amazonaws.com/sdformat/api/3.0/classsdf_1_1Element.html#a29fa74bb6bf29701e82b6353e732d31d).
+Currently, this doesn't represent a problem since also the plugins are reading the configurations with `GetElement()`.
+
+### Usage
+Add the ConfigurationOverride plugin to the model sdf before the overriden plugin declaration (i.e. before the `<include>`) and define the `plugin_name` as the `name` of the plugin that will be modified. 
+The properties of the ConfigurationOverride plugin will be then used in order to override the properties of the desired plugin.
+
+e.g. For changing the `<initialConfiguration>` of the `ControlBoard` (`controlboard_right_arm`), the model will be defined as follow:
```
traversaro: Minor: Note that `gazebo-yarp-plugins` is completly independent from `icub-gazedo` or `icub-models`, so perhaps is is more clear if you explicitly say start the sentence as : "e.g. If you are using the icub gazebo model from [`icub-models`](https://github.com/robotology/icub-models), and you want to change the `<initialConfiguration>` etc etc" .

lrapetti: Right, thanks! I have changed the sentence and rebased

gabrielenava:
I tested the plugin for the [icub on seesaw](https://github.com/robotology/icub-gazebo-wholebody/tree/master/worlds/icub_seesaw_world) and [icub standup](https://github.com/robotology/icub-gazebo-wholebody/tree/master/worlds/icub_standup_world) models. In particular, these models were previously loaded by means of a `Gazebo world` file defined as follows:

```[xml]
<?xml version="1.0"?>
<sdf version="1.4">
  <world name="default">

    <!-- Light -->
    <include>
      <uri>model://sun</uri>
    </include>

    <!-- Ground Plane -->
    <include>
      <uri>model://ground_plane</uri>
    </include>

    <!-- Seesaw -->
    <include>
        <uri>model://seesaw</uri>
    </include>

    <!-- iCub above the seesaw -->
    <include>
    <uri>model://icub_on_seesaw</uri>
    </include>

  </world>
</sdf>
```

Where the `icub_on_seesaw` model was a duplicate of the original `icub_no_hands` model, with a different joints configuration. I modified the worlds files as follows:

```[xml]
<?xml version="1.0"?>
<sdf version="1.4">
  <world name="default">

    <!-- Light -->
    <include>
      <uri>model://sun</uri>
    </include>

    <!-- Ground Plane -->
    <include>
      <uri>model://ground_plane</uri>
    </include>

    <!-- Seesaw -->
    <include>
        <uri>model://seesaw</uri>
    </include>

    <!-- iCub above the seesaw -->
    <model name="iCub">
      <plugin name='torso_configuration_override' filename='libgazebo_yarp_configurationoverride.so'>
        <yarpPluginConfigurationOverride plugin_name='controlboard_torso'> </yarpPluginConfigurationOverride>
        <initialConfiguration> 0 0 0.0824</initialConfiguration>
      </plugin>
      <plugin name='larm_configuration_override' filename='libgazebo_yarp_configurationoverride.so'>
        <yarpPluginConfigurationOverride plugin_name='controlboard_left_arm_no_arm'> </yarpPluginConfigurationOverride>
        <initialConfiguration>-0.6278 0.5231 0.0010 0.8727 0 0 0</initialConfiguration>
      </plugin>
      <plugin name='rarm_configuration_override' filename='libgazebo_yarp_configurationoverride.so'>
        <yarpPluginConfigurationOverride plugin_name='controlboard_right_arm_no_arm'> </yarpPluginConfigurationOverride>
        <initialConfiguration>-0.6278 0.5231 0.0010 0.8727 0 0 0</initialConfiguration>
      </plugin>
      <plugin name='lleg_configuration_override' filename='libgazebo_yarp_configurationoverride.so'>
        <yarpPluginConfigurationOverride plugin_name='controlboard_left_leg'> </yarpPluginConfigurationOverride>
        <initialConfiguration>0.2094 0.0873 0 -0.1745 0.0479 -0.0898</initialConfiguration>
      </plugin>
      <plugin name='rleg_configuration_override' filename='libgazebo_yarp_configurationoverride.so'>
        <yarpPluginConfigurationOverride plugin_name='controlboard_right_leg'> </yarpPluginConfigurationOverride>
        <initialConfiguration>0.2094 0.0873 0 -0.1745 0.0479 -0.0898</initialConfiguration>
      </plugin>
      <include>
        <uri>model://icub</uri>
        <pose>-0.05 0 0.6926 0 0 3.14</pose>
      </include>
    </model>

  </world>
</sdf>
```

Now, the model loaded in the world file is the original `icub_no_hands` model, and the configuration is overwritten by means of the ConfigurationOverride plugin. I also removed the `icub_on_seesaw` model from the [models](https://github.com/robotology/icub-gazebo-wholebody/tree/master/models) folder. I did the same procedure for `icub_on_a_chair`. 

I tested the plugin on Gazebo. To be sure I was not using anymore the previous configuration I removed my INSTALL folder and I installed the iCub models again. Here are the results of running Gazebo + `icub seesaw/standup world file`:

screenshot from 2018-11-21 11-48-48
screenshot from 2018-11-21 11-49-10

So the plugin is working fine. I created a [devel](https://github.com/robotology/icub-gazebo-wholebody/tree/devel) branch in `icub-gazebo-wholebody` as requested in [this issue](https://github.com/robotology/icub-gazebo-wholebody/issues/21).

traversaro:
Thanks @lrapetti and @gabrielenava . This is a feature that deserves some recognition, we do not mantain a changelog for gazebo-yarp-plugins unfortunatly (any volunteer for doing that is highly welcome : ) ) but we should remember to mention this when we will announce the next release.

lrapetti:
Great! Thanks @gabrielenava and @traversaro

PR #403, opened 2018-11-21
Feature/virtual analog wrapper

prashanthr05:
This PR implements the server side implementation of the virtual analog sensor interface in the control board plugin.

In order to use this interface, the user should add a flag `useVirtualAnalogSensor` to the top level of .ini (yarpConfigurationFile) and set this flag to true. By default, this flag is set to false. 

To open the interface, a new parameters group similar to `[WRAPPER]` should be defined in the .ini file.
This group should be named `[VIRTUAL_ANALOG_SERVER]`. It should be defined with the following parameters,
- `device` parameter should be set to `virtualAnalogServer`
- `name` parameter is the name of the `remote` YARP port corresponding to the virtual analog sensor device opened in the client side.    
- `networks` parameter is the list of motor control device to use. However, this parameter is already declared under the `[WRAPPER]` group and is fetched from there.
- A corresponding mapping list for the opened motor control device must also be specified. This is also available in the `[WRAPPER]` group.
As an example,the .ini file may look like this
```ini
useVirtualAnalogSensor true

[WRAPPER]
** wrapper device related parameters**
networks ( left_leg )
left_leg 0 5 0 5

[VIRTUAL_ANALOG_SERVER]
device virtualAnalogServer
name /<robot_name>/joint_vsens/left_leg:i 
```

If opened successfully, the joint torques which are usually obtained from the Gazebo's physics engine are initialized to zero. And this server waits for the client side devices to publish the measurements to be copied into the virtual analog sensor values.  Only when the client is open and a proper connection is made, meaningful values are obtained through this interface. In case, there is more than one client device, we will need a remapper as well, to connect all the client devices into one.

We need to make sure that the client side already has a virtual analog client device opened and a virtual analog remapper to remap these devices onto a single device. An example implementation of these devices can be found in [codyco-modules/virtualAnalogRemapper](https://github.com/robotology/codyco-modules/tree/master/src/devices/virtualAnalogRemapper) and [codyco-modules/virtualAnalogClient](https://github.com/robotology/codyco-modules/tree/master/src/devices/virtualAnalogClient).


plugins/controlboard/src/ControlBoard.cc
```diff
@@ -68,29 +79,29 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
         // Add the gazebo_controlboard device driver to the factory.
         yarp::dev::Drivers::factory().add(new yarp::dev::DriverCreatorOf<yarp::dev::GazeboYarpControlBoardDriver>("gazebo_controlboard", "controlboardwrapper2", "GazeboYarpControlBoardDriver"));
 
-        // Getting .ini configuration file parameters from sdf
+        //Getting .ini configuration file from sdf
         bool configuration_loaded = GazeboYarpPlugins::loadConfigModelPlugin(_parent, _sdf, m_parameters);
-
+
```
traversaro: Please remove the trailing whitespace/non essential changes, thanks!

plugins/controlboard/src/ControlBoardDriverVirtualAnalogSensor.cpp
```diff
@@ -0,0 +1,68 @@
...
+namespace yarp {
+    namespace dev {
+      
+
+        VAS_status GazeboYarpControlBoardDriver::getVirtualAnalogSensorStatus(int ch)
+        {
+            if (ch < 0 || ch >= this->getVirtualAnalogSensorChannels())
+            {
+                yError() << "GazeboYarpControlBoardDriver: VirtualAnalogServer: getState failed: requested channel " << ch << 
+                            "while the client is configured wtih " << this->getVirtualAnalogSensorChannels() << "channels";
```
traversaro: Fora next PR: `wtih` --> `with`.

Issue #402, opened 2018-11-19
World interface plugin: makeBox


valegagge:
I' m using the rpc port of world interface plugin to create a box with a particular name (for example myBox), but the plugin creates a box called myBoxN, where N is the current number of box in the world.

please see code here:
https://github.com/robotology/gazebo-yarp-plugins/blob/154f236e16a18bcc5ef9a9dd26d1a4c8792d8cac/plugins/worldinterface/src/worldproxy.cpp#L260

Could someone tell me if there is an motivation behind this behaviour? (This happens both on matser and on devel branch)
Thanks a lot

valegagge:
MakeSphere has the same behaviour, but I needed to changed it according to @randaz81 in order that makesphere creates a sphere with the given name.

please see https://github.com/robotology/gazebo-yarp-plugins/commit/de34ac26bcbd8565d1d55875d7d7c2ec34b93302

In my opinion, any makeObject function with ObjectName as parameter should create an object with the given name.

Since this issue is not blocking for my activities, I prefer to open an issue and ask if this behaviour is desidered before doing a pull request.

So, I wait their comments or comment of MakeBox users :)

lornat75:
I can't remember why it was implemented in this way, the behavior you describe makes sense to me.

PR #404, opened 2018-11-21
update worldinterface::makeBox

valegagge:
worldinterface plugin: now the makeBox function doesn't append number at the end of given name of box.

please see https://github.com/robotology/gazebo-yarp-plugins/issues/402

### DIFF
--- a/plugins/worldinterface/src/worldinterfaceserverimpl.h
+++ b/plugins/worldinterface/src/worldinterfaceserverimpl.h
@@ -26,7 +26,7 @@ public:
      * @param radius radius of the sphere [m]
      * @param pose pose of the sphere [m]
      * @param color color of the sphere
-     * @return returns a string that contains the name of the object in the world
+     * @return returns a string that contains the name of the object in the world or empty string on error
      */
     virtual std::string makeSphere(const double radius, const GazeboYarpPlugins::Pose& pose, const GazeboYarpPlugins::Color& color, const std::string& frame_name = "", const std::string& object_name = "",const bool gravity_enable = 0, const bool collision_enable = 1);
     /**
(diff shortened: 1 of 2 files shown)


valegagge:
@traversaro also makeFrame, right?

traversaro:
Indeed, I missed that!

valegagge:
Done!

traversaro:
Ok for me, but let's wait for @randaz81 opinion given that is the one more involved in this.

traversaro:
I just saw that @randaz81 is away till the 4th of December. If you think he will not have any problem with this, I think we can merge.

valegagge:
I'm not used to work with gazebo and in particular with worldinterface plugin. 
I think the current behaviour of mentioned functions doesn't seem usefull.
I prefer tomerge and in case of issue we can restore the current behaviour, but , since this pr is not blocking for me, I can wait @randaz

traversaro:
friendly ping @randaz81

randaz81:
@valegagge it seems ok to me, but can you please do this additional check?
What happens if a user tries to create an object with the same name of an object that already exists? Previously, a number was appended, resulting in a different name. Now I think that two objects with the same name are created. Maybe this action should be prevented an error message displayed?

valegagge:
@randaz81 
>  it seems ok to me, but can you please do this additional check?

Done! please see [this commit](https://github.com/robotology/gazebo-yarp-plugins/pull/404/commits/5ccd19d74eb5442df0a99dc3427a1a773d36d685)

About the error message displayed, I used the existing error management: prints yError << "error_string" and returns an empty string. This means that on error the response of rpc command is nothing and only on logger there is the error or on bash where gazebo is running.

Issue #407, opened 2019-01-16
Invalid camera parameters in gazebo_depthCamera plugin


jchevrie:
The values of `focalLengthX` and `focalLengthY` returned by the plugin (when asking the intrinsic parameters of the camera through rpc via `visr get intp`) seem to be expressed in meters, while they are expressed in pixel size in the case of hardware components (for example the realsense2Driver in YARP).

This should be fixed so that the simulation provides the same interface as the hardware, or was there a reason behind this behavior?

Nicogene:
Good catch, @traversaro is right, in [realsense documentation](https://github.com/IntelRealSense/librealsense/wiki/Projection-in-RealSense-SDK-2.0#intrinsic-camera-parameters) there is written that `fx` and `fy` are represented as a multiple of pixel width and height.

It is a bug inside a `realsense2` driver, a conversion should be done in order to have it in mm.

traversaro:
If in gazebo-yarp-plugins `focalLengthX` is reported in meters, we need also to fix it there to be consistent with the YARP interface.

Nicogene:
Issue opened in YARP  robotology/yarp#1949

jchevrie:
I agree that we should follow the documentation if any implementation of `yarp::dev::IRgbVisualParams::getRgbIntrinsicParam` is currently following it.
I'm wondering if it is a bug in the RealSense driver or a mistake in the documentation?

If I'm not mistaken, having the focal length in mm (or meters) does not allow the conversion from 3D metric space to 2D pixel space using only the intrinsic parameters provided by `yarp::dev::IRgbVisualParams::getRgbIntrinsicParam` (https://github.com/robotology/yarp/blob/47e8ae58889789465b15d90f85e1863aeab4c2c5/src/libYARP_dev/include/yarp/dev/IVisualParams.h#L155).
For this a pixel/meter ratio or the physical size of the pixels is missing in the list.
Having the focal length directly expressed in pixels (as is the case with the RealSense) made more sense to me since it solves this issue.

traversaro:
>  I'm wondering if it is a bug in the RealSense driver or a mistake in the documentation?

That is a good point. If you can find an agreement with the maintainers of the interfaces and of the existing driver implementations (that I guess are @Nicogene and @barbalberto) and fix the YARP documentation, I will then be happy to merge the proper fix in gazebo-yarp-plugins.

claudiofantacci:
[...]

PR #408, opened 2019-01-16
Fix invalid camera parameters in gazebo_depthCamera plugin

jchevrie:
Fix for issue #407 

Fix `focalLengthX` and `focalLengthY` values returned by the plugin when asking the intrinsic parameters of the camera through rpc via the command `visr get intp`.
The new returned values are now pixel/meter ratios instead of meters, to be similar to the interface of the hardware cameras.

### DIFF
--- a/plugins/depthCamera/src/DepthCamera.cc
+++ b/plugins/depthCamera/src/DepthCamera.cc
@@ -94,7 +94,7 @@ void GazeboYarpDepthCamera::Load(sensors::SensorPtr _sensor, sdf::ElementPtr _sd
     }
 
     //Open the driver
-    //Force the device to be of type "gazebo_forcetorque" (it make sense? probably yes)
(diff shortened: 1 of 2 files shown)


traversaro:
Given the fixes in https://github.com/robotology/yarp/pull/1960, how this PR should be updated (if it needs to be updated at all)?

jchevrie:
In the end we chose pixels as unit instead of meters, so this fix is still relevant.

I updated the code by adding the new mandatory parameter `physFocalLength` that was introduced.
However I set the returned value to 0.0 (meaning invalid) because I don't think we set it in gazebo for now (it has a default value 1.0 which should probably not be used as is).

traversaro:
Great, thanks a lot @jchevrie . If @Nicogene or @barbalberto could review this PR given their expertise on this interface it would be great, thanks!

barbalberto:
Merged, thanks @jchevrie for your work on this topic

PR #412, opened 2019-02-08
Fix bug in deleteObject of WorldInterface plugin

valegagge:
Due to this bug, the function deleted alway the first object in its list and not object with the name specified in the parameter of function


valegagge:
cc @randaz81

PR #416, opened 2019-03-08
Change behavior of method GazeboYarpControlBoardDriver::checkMotionDone so that it adheres to that of the real control board

### LINKED ISSUE #377: Wrong implementation of  yarp::dev::IPositionControl::checkMotionDone in Control board plugin
It seems that the implementation of the method
`
yarp::dev::IPositionControl::checkMotionDone (const int n_joint, const int * joints, bool * flags)
`
within the control board plugin is wrong.

According to the [documentation](http://www.yarp.it/classyarp_1_1dev_1_1IPositionControl.html#a78e3171536e8a6f0705435f9d3baf769) `flags` is a pointer to the return value that should be the logical "and" of all set of joints. However the code

https://github.com/robotology/gazebo-yarp-plugins/blob/3fa2c45a4b8a63ef15c524462124b505f29b3ed9/plugins/controlboard/src/ControlBoardDriverPositionControl.cpp#L235

tries to write into the variable `flags` as it was an array of booleans resulting in a

```
*** stack smashing detected ***: gzserver terminated
```

when the pointer to a single boolean is given to the function.

xEnVrE:
This PR fixes the implementation of the method

`
yarp::dev::IPositionControl::checkMotionDone (const int n_joint, const int * joints, bool * flags)
`
within the control board plugin as described in the issue #377.

Fixes #377

### DIFF
--- a/plugins/controlboard/src/ControlBoardDriverPositionControl.cpp
+++ b/plugins/controlboard/src/ControlBoardDriverPositionControl.cpp
@@ -218,12 +218,15 @@ bool GazeboYarpControlBoardDriver::relativeMove(const int n_joint, const int *jo
     return ret;
 }
 
-bool GazeboYarpControlBoardDriver::checkMotionDone(const int n_joint, const int *joints, bool *flags)
+bool GazeboYarpControlBoardDriver::checkMotionDone(const int n_joint, const int *joints, bool *flag)
 {
-    if (!joints || !flags) return false;
+    if (!joints || !flag) return false;
     bool ret = true;
+    *flag = true;
     for (int i = 0; i < n_joint && ret; i++) {
-        ret = checkMotionDone(joints[i], &flags[i]);
+        bool done;
+        ret = checkMotionDone(joints[i], &done);
+        (*flag) &= done;
     }
     return ret;
 }


plugins/controlboard/src/ControlBoardDriverPositionControl.cpp
```diff
@@ -222,8 +222,11 @@ bool GazeboYarpControlBoardDriver::checkMotionDone(const int n_joint, const int
 {
     if (!joints || !flags) return false;
     bool ret = true;
+    *flags = true;
     for (int i = 0; i < n_joint && ret; i++) {
-        ret = checkMotionDone(joints[i], &flags[i]);
+	bool done;
```
traversaro: Can you use 4 spaces instead of `tabs` ? Thanks!

traversaro:
Can you target the `master` branch, as this is a bug fix? Thanks!

plugins/controlboard/src/ControlBoardDriverPositionControl.cpp
xEnVrE: The editor I use is configured not to use `tabs` at all. I will check carefully again.

xEnVrE:
Changes done and moved target branch to `master`.

traversaro:
Thanks a lot, waiting for Travis and then we will merge.

traversaro:
Thanks!

Issue #417, opened 2019-03-27
Limitations of externalWrench plugin


Yeshasvitvs:
[externalWrench](https://github.com/robotology/gazebo-yarp-plugins/tree/master/plugins/externalwrench) plugin has a couple of shortcomings:

- [x] There is no error message if the link is not discovered. It only acknowledges if the command format is correctly used i.e `Response: "[ACK] Correct command format"` is shown. The expected command format is `[link] [force] [torque] [duration]`. Any string used in place of `[link]` will give the correct command response instead of giving an error that the link is not discovered.

- [x] Secondly, the `[link]` string is expected to be a full scoped link name of the model spawned in gazebo e.e `iCub::l_hand`. But if the model is spawned with a higher levels of scope like `iCub::iCub::l_hand` the external wrench plugin doesn't have any effect. This behaviour is checked on both @lrapetti machine and my machine. Some of the recent worlds in icub-gazebo-wholebody repo like [icub_standup_world](https://github.com/robotology/icub-gazebo-wholebody/blob/devel/worlds/icub_standup_world/icub_standup.world) uses [include](https://github.com/robotology/icub-gazebo-wholebody/blob/devel/worlds/icub_standup_world/icub_standup.world) tag for including the model in the world file. This is the reason for the addition scope of `iCub` resulting in the final link scope name as `iCub::iCub::l_hand`. So, it is better to have the command format as `[model] [link] [force] [torque] [duration]` and the user has to specify the first model name in the scope e.g `iCub` and the link name `l_hand`. This kinda functionality is already present in the [linkattacher](https://github.com/robotology/gazebo-yarp-plugins/tree/master/plugins/linkattacher) plugin. 

- [x] Finally, the plugin can be used to genereate only one instance of the external wrench and currently it is not possible to use multiple external wrenches. Previously a PR https://github.com/robotology/gazebo-yarp-plugins/pull/293 was opened to add the functionality of multiple external wrenches. We can update the [externalWrench](https://github.com/robotology/gazebo-yarp-plugins/tree/master/plugins/externalwrench) to have this additional feature. 

@traversaro Please let me know your comments

traversaro:
I agree with all remarks.

Yeshasvitvs:
[...]

PR #418, opened 2019-03-28
Update external wrench plugin with mutiple wrenches and improve rpc responses

Yeshasvitvs:
Following the limitations of the [externalWrench](https://github.com/robotology/gazebo-yarp-plugins/tree/master/plugins/externalwrench) documented in https://github.com/robotology/gazebo-yarp-plugins/issues/417 this PR updated the plugin with new functionalities.


traversaro:
Can we find a reviewer in the lab? (Perhaps @lrapetti?) I am a bit short of time till next week.

plugins/externalwrench/include/gazebo/ApplyExternalWrench.hh
```diff
@@ -102,6 +96,7 @@ private:
     msgs::Visual            m_visualMsg;
 
     bool                    m_newCommand;
+    bool                    m_setMessage;
```
lrapetti: unused variable `m_setMessage`

plugins/externalwrench/include/gazebo/ApplyExternalWrench.hh
```diff
@@ -56,41 +62,29 @@ private:
...
-        double            duration;
-    };
+    bool getCandidateLink(physics::LinkPtr& candidateLink, std::string candidateLinkName);
 
     /// \brief Robot name that will be used to open rpc port
     std::string          robotName;
-    double               timeIni;
-//     yarp::os::Bottle     bufferBottle;
 
+    double               timeIni;
```
lrapetti: this variable seems to be unused since the duration check has been moved to ExternalWrench.cc.

Also, I think `#include <yarp/os/Time.h>` could be safely removed.

plugins/externalwrench/include/gazebo/ApplyExternalWrench.hh
```diff
@@ -56,41 +62,29 @@ private:
...
     yarp::os::Network       m_yarpNet;
     RPCServerThread         m_rpcThread;
     yarp::os::Property      m_iniParams;
 
     physics::ModelPtr       m_myModel;
-    /// \brief Link on which the wrench will be applied
-    std::string             m_modelScope;
-    std::string             m_subscope;
-    std::string             m_linkName;
+
```
lrapetti: extra blank line

plugins/externalwrench/src/ApplyExternalWrench.cc
```diff
@@ -310,19 +115,46 @@ void RPCServerThread::run()
...
+
+                }
+
+                //Creating new instances of external wrenches
+                boost::shared_ptr<ExternalWrench> newWrench(new ExternalWrench);
+                if(newWrench->setWrench(m_robotModel, m_cmd))
+                {
+                    wrenchesVectorPtr->push_back(newWrench);
+                }
+                else yError() << "Failed to set new wrench values!";
```
lrapetti: Here it should be possible to reply with an `rpc` error message as [done below](https://github.com/robotology/gazebo-yarp-plugins/pull/418/files#diff-6f34d47dcd5558cbef158050b5563c76R158):
```
this->m_reply.clear();
this->m_reply.addString ( "ERROR: unable to get the desired link" );
this->m_rpcPort.reply ( this->m_reply );
```

plugins/externalwrench/src/ExternalWrench.cc
```diff
@@ -0,0 +1,181 @@
...
+}
+
+bool ExternalWrench::setWrench(physics::ModelPtr& _model,yarp::os::Bottle& cmd)
+{
+    model = _model;
+
+    //get link name from command
+    std::string linkName = cmd.get(0).asString();
+    if (linkName.find_last_of("::") != 0) {
+        wrenchPtr->link_name = linkName.substr(linkName.find_last_of("::")+1); //case of scoped link name
```
lrapetti: Probably this is not needed, we can simply use `wrenchPtr->link_name = linkName` also in case of scoped `linkName`.

If a scoped link name is passed we don't want to cancel submodel names which can be useful when multiple models are used (e.g. they exists both `iCub::iCub_1::l_hand` and `iCub::iCub_2::l_hand`). [hasEnding](https://github.com/robotology/gazebo-yarp-plugins/pull/418/files#diff-bdf55637a7d0134a9ccb12f75775a362R36) should handle also scoped names.

**e.g.**
To make a practical example, if we have to models with `l_hand`, we could have two links with names `iCub::iCub_1::l_hand` and `iCub::iCub_2::l_hand`.
If now we want to target the hand of `iCub::iCub_2`, we should be able to use the scoped name `iCub_2::l_hand`. 
But the way it is implemented now it will just search for `l_hand` ending with `hasEnding` which could result in the other model link `iCub::iCub_2::l_hand`. 
If instead we simply use the scoped name as `linkname`, when `hasEnding` will be called, the correct link `iCub::iCub_2::l_hand` would be found.

Yeshasvitvs: > But the way it is implemented now it will just search for l_hand ending with hasEnding which could result in the other model link iCub::iCub_2::l_hand.

I believe this will not happen because this plugin is a model plugin. So, it will only consider the model to which it is attached.

> Probably this is not needed, we can simply use wrenchPtr->link_name = linkName also in case of scoped linkName.

You are correct. I put this check in place to ensure back compatibility for current users who use command format link as `iCub::l_hand`. But the `hasEnding` function will take care of that.

plugins/externalwrench/include/gazebo/ApplyExternalWrench.hh
Yeshasvitvs: Done

plugins/externalwrench/include/gazebo/ApplyExternalWrench.hh
Yeshasvitvs: Done

plugins/externalwrench/include/gazebo/ApplyExternalWrench.hh
Yeshasvitvs: Done

plugins/externalwrench/src/ApplyExternalWrench.cc
Yeshasvitvs: Updated the response handling

plugins/externalwrench/src/ExternalWrench.cc
Yeshasvitvs: Done

lrapetti:
Looks Good To Me, 
@traversaro I think we can merge.

lrapetti:
It would be better to also create a `README.md` for the `externawrench` plugin (for example like the one that can be found for the [`linkattacher`](https://github.com/Yeshasvitvs/gazebo-yarp-plugins/blob/feature/update_externalWrench/plugins/linkattacher/README.md)).

traversaro:
I have added a few comments to the PR as well. For future PR, can you use human understandable PR titles, in place of the one automatically generated by github from the branch names? Thanks!

plugins/externalwrench/src/ExternalWrench.cc
```diff
@@ -0,0 +1,173 @@
...
+    color[2] = static_cast <float> (rand()) / static_cast <float> (RAND_MAX);
+    color[3] = static_cast <float> (rand()) / static_cast <float> (RAND_MAX);
+
+    wrenchPtr->link_name = "world";
+    wrenchPtr->force.resize(3,0);
+    wrenchPtr->torque.resize(3,0);
+    wrenchPtr->duration = 0.0;
+
+    count++;
+    tick = yarp::os::Time::now();
```
traversaro: I am afraid that in general this  will not give you the expected duration of the external force, unless `gzserver` is executed with the `YARP_CLOCK=/clock` env variable and with the `gazebo_yarp_clock` plugin. If one of this conditions is not true, `yarp::os::Time::now()` will return you the system time, resulting in a duration of the external force in simulation that may be much shorter than expected (depending on the RTF) and even worse a duration that is not reproducible.

traversaro: The proper solution to this is to add a `const gazebo::common::UpdateInfo& _info` argument to the applyWrenches method, and use it to retrieved the simulation time step. See other onUpdate methods in the repo to get an inspiration https://github.com/robotology/gazebo-yarp-plugins/search?q=onUpdate&unscoped_q=onUpdate .

plugins/externalwrench/include/gazebo/ApplyExternalWrench.hh
```diff
@@ -27,22 +29,26 @@ public:
...
     void                setRobotName(std::string robotName);
-    void                setScopedName(std::string scopedName);
-    void                setDefaultLink(const std::string& defaultLink);
-    void                setDurationBuffer(double d);
-    double              getDurationBuffer();
+    void                setRobotModel(physics::ModelPtr robotModel);
     void                setNewCommandFlag(std::int32_t flag);
     virtual void        onStop();
+
+    boost::shared_ptr< std::vector< boost::shared_ptr<ExternalWrench>>> wrenchesVectorPtr{new std::vector< boost::shared_ptr<ExternalWrench>>()};
```
traversaro: Why can't we use a good old `std::vector<ExternalWrench>`?

plugins/externalwrench/include/gazebo/ExternalWrench.hh
```diff
@@ -0,0 +1,83 @@
...
+
+#include <boost/lexical_cast.hpp>
+
+using namespace gazebo;
+
+class ExternalWrench
+{
+private:
+
+   static int count;
```
traversaro: Everything seems to be ok from what I understand, but static variables are always a bit tricky do deal with. 
I don't think this needs to block the PR, but I think you can obtain something similar by setting the `count` to the in ExternalWrench constructor as you build it in the plugin class.

plugins/externalwrench/src/ApplyExternalWrench.cc
```diff
@@ -310,28 +115,90 @@ void RPCServerThread::run()
...
+        } else{
+            if((command.size() == 8) && (command.get(0).isString() \
+                && ( command.get ( 1 ).isDouble() || command.get ( 1 ).isInt() )  && ( command.get ( 2 ).isDouble() || command.get ( 2 ).isInt() ) && ( command.get ( 3 ).isDouble() || command.get ( 3 ).isInt() ) \
+                && ( command.get ( 4 ).isDouble() || command.get ( 4 ).isInt() )  && ( command.get ( 5 ).isDouble() || command.get ( 5 ).isInt() ) && ( command.get ( 6 ).isDouble() || command.get ( 6 ).isInt() )  \
+                && ( command.get ( 7 ).isDouble() || command.get ( 7 ).isInt() )) ) {
+                this->m_message = "[ACK] Correct command format";
                 m_lock.lock();
                 // new-command flag
-                command.addInt32(1);
+                command.addInt(1);
```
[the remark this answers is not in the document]
Yeshasvitvs: > not the suggested method to use

**not** the suggested method or **is** the suggested method?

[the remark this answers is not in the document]
traversaro: Sorry, *is* the suggested method, I fixed the comment.

Yeshasvitvs:
> It would be better to also create a `README.md` for the `externawrench` plugin (for example like the one that can be found for the [`linkattacher`](https://github.com/Yeshasvitvs/gazebo-yarp-plugins/blob/feature/update_externalWrench/plugins/linkattacher/README.md)).

README file added https://github.com/robotology/gazebo-yarp-plugins/pull/418/commits/7366f2fba071c333d7ade131f6e1a02b4cb43c73

plugins/externalwrench/include/gazebo/ExternalWrench.hh
Yeshasvitvs: @traversaro I am not following you exactly. The `count` variable is actually increased in the constructor of `ExternalWrench` https://github.com/robotology/gazebo-yarp-plugins/blob/f796908d68c22d046586b4248a9fc6bf79b1190d/plugins/externalwrench/src/ExternalWrench.cc#L22

plugins/externalwrench/src/ExternalWrench.cc
```diff
@@ -0,0 +1,173 @@
...
+    duration_done = false;
+}
+
+bool ExternalWrench::getLink()
+{
+    // Get the exact link with only link name instead of full_scoped_link_name
+    links = this->model->GetLinks();
```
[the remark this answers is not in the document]
Yeshasvitvs: Done

plugins/externalwrench/include/gazebo/ExternalWrench.hh
```diff
@@ -0,0 +1,83 @@
...
+        yarp::sig::Vector        torque;
+        double                   duration;
+   };
+
+   std::unique_ptr<wrenchCommand> wrenchPtr{new wrenchCommand()};
+
+   double                        tick;
+   double                        tock;
+
+   bool                          model_has_link;
```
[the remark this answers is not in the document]
Yeshasvitvs: Done

plugins/externalwrench/include/gazebo/ExternalWrench.hh
```diff
@@ -0,0 +1,83 @@
...
+   physics::ModelPtr             model;
+   physics::LinkPtr              link;
+   physics::Link_V               links;
+
+   transport::NodePtr            m_node;
```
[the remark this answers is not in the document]
Yeshasvitvs: Done, remove `m_` prefix

plugins/externalwrench/src/ExternalWrench.cc
```diff
@@ -0,0 +1,173 @@
+#include "ExternalWrench.hh"
+
+int ExternalWrench::count = 0;
+
+//Initializing wrench command
+ExternalWrench::ExternalWrench()
+{
+    model_has_link = false;
+
+    //srand(boost::lexical_cast<float>(std::time(NULL)));
```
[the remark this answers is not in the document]
Yeshasvitvs: Done

plugins/externalwrench/src/ApplyExternalWrench.cc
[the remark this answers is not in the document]
Yeshasvitvs: Done

plugins/externalwrench/include/gazebo/ExternalWrench.hh
```diff
@@ -0,0 +1,83 @@
+#ifndef EXTERNALWRENCH_H
```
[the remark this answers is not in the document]
Yeshasvitvs: Done

Yeshasvitvs:
> Some comments.
> Bonus points: what happens if the simulation is reset while a force is being applied? See
> 
> [gazebo-yarp-plugins/plugins/controlboard/src/ControlBoardDriver.cpp](https://github.com/robotology/gazebo-yarp-plugins/blob/01ad1d1252b0827033fd57058f066cc790b893c6/plugins/controlboard/src/ControlBoardDriver.cpp#L341)
> 
> Line 341 in [01ad1d1](/robotology/gazebo-yarp-plugins/commit/01ad1d1252b0827033fd57058f066cc790b893c6)
> 
>  this->m_resetConnection = 
> for an example on how to deal with the reset event.

On resetting gazebo, the wrenches are still applied as provided by the user commands. Do you think the wrenches should be cleared ?

plugins/externalwrench/src/ExternalWrench.cc
```diff
@@ -0,0 +1,173 @@
...
+
+void ExternalWrench::setVisual()
+{
+    //Wrench Visual
+    m_node = transport::NodePtr(new gazebo::transport::Node());
+    this->m_node->Init(model->GetWorld()->Name());
+    m_visPub = this->m_node->Advertise<msgs::Visual>("~/visual",100);
+
+    // Set the visual's name. This should be unique.
+    std::string visual_name = "__" + wrenchPtr->link_name + "__CYLINDER_VISUAL__" + boost::lexical_cast<std::string>(count);
```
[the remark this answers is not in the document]
Yeshasvitvs: Done

plugins/externalwrench/src/ExternalWrench.cc
```diff
@@ -0,0 +1,173 @@
...
+
+int ExternalWrench::count = 0;
+
+//Initializing wrench command
+ExternalWrench::ExternalWrench()
+{
+    model_has_link = false;
+
+    //srand(boost::lexical_cast<float>(std::time(NULL)));
+    srand(rand()%100);
```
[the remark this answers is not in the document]
Yeshasvitvs: >  either you initialize a local random generator that you seed with count for getting the color components

@traversaro what you mean by "local random generator" ?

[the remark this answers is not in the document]
traversaro: Something like:
~~~
std::mt19937 gen(42);
std::uniform_real_distribution<> dis(0.0, 1.0);
for(int i=0; i < 10; i++)
{
    std::cerr << "Random " << i << " : " << dis(gen) << std::endl;
}
~~~
Given that you  seed it with a fixed number (42) the output of this function it is deterministic.

traversaro:
> On resetting gazebo, the wrenches are still applied as provided by the user commands. Do you think the wrenches should be cleared? @traversaro

My understanding of reset is that the simulation should restart from the initial condition. If a user asked for an external force of 1 N on one link for one hour, I think after a reset, the force should be cleared. Think for example of an RL scenario: would it make sense to keep the old force?

plugins/externalwrench/include/gazebo/ExternalWrench.hh
```diff
@@ -0,0 +1,83 @@
...
+   float      color[4];
+   struct wrenchCommand
+   {
+        std::string              link_name;
+        yarp::sig::Vector        force;
+        yarp::sig::Vector        torque;
+        double                   duration;
+   };
+
+   std::unique_ptr<wrenchCommand> wrenchPtr{new wrenchCommand()};
```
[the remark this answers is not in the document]
Yeshasvitvs: @traversaro I tried to use without pointers 

> > On resetting gazebo, the wrenches are still applied as provided by the user commands. Do you think the wrenches should be cleared? @traversaro
> 
> My understanding of reset is that the simulation should restart from the initial condition. If a user asked for an external force of 1 N on one link for one hour, I think after a reset, the force should be cleared. Think for example of an RL scenario: would it make sense to keep the old force?

Agreed!

Yeshasvitvs:
> > On resetting gazebo, the wrenches are still applied as provided by the user commands. Do you think the wrenches should be cleared? @traversaro
> 
> My understanding of reset is that the simulation should restart from the initial condition. If a user asked for an external force of 1 N on one link for one hour, I think after a reset, the force should be cleared. Think for example of an RL scenario: would it make sense to keep the old force?

Agreed!

plugins/externalwrench/include/gazebo/ExternalWrench.hh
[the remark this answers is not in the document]
Yeshasvitvs: > I am a bit confused. Why are you using `std::unique_ptr` here?
> Why not a simple attribute of the class:
> 
> ```c++
>   wrenchCommand wrench;
> ```
> 
> or even better, given that the structure `wrenchCommand` is not used elsewhere, an anonymous struct:
> 
> ```c++
>    struct
>    {
>         std::string              link_name;
>         yarp::sig::Vector        force;
>         yarp::sig::Vector        torque;
>         double                   duration;
>    } wrench;
> ```
> 
> or given that the structure is never used standalone, why not simply including the attributes of the structure as attributes of the class?

@traversaro I remember back in time starting out with simple data structures without pointers but then I quickly ran into some problems while handling dynamic wrench addition and removal in multiple wrench operation mode. That is why I belive I migrated to using pointer to handle them correctly. Now, I tested to change back to simple data structures again and in the multiple wrench case the visuals are funny looking 

ezgif com-video-to-gif

But with the pointers usage it is stable

ezgif com-video-to-gif (1)

At the moment I prefer not to investigate it further and I would like to stick to the pointers usage.

plugins/externalwrench/include/gazebo/ApplyExternalWrench.hh
Yeshasvitvs: Similar to https://github.com/robotology/gazebo-yarp-plugins/pull/418#discussion_r273440396

plugins/externalwrench/include/gazebo/ExternalWrench.hh
traversaro: Yes, but what happens if there are two different externalwrench plugins in the same model? As far as I was able to see, everything should work fine, but keeping `count` as a static variable may lead to some subtle bugs if this code is modified in the future.

plugins/externalwrench/src/ExternalWrench.cc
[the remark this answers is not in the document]
Yeshasvitvs: Updated color random generation https://github.com/robotology/gazebo-yarp-plugins/pull/418/commits/a4323debf42c5449f14ed5f2cfcc8880a1319a24

ezgif com-video-to-gif (2)

plugins/externalwrench/include/gazebo/ExternalWrench.hh
Yeshasvitvs: >  what happens if there are two different externalwrench plugins in the same model? 

At the moment this is not possible to do by a user as there will be port address conflict while trying to open the rpc port with the model name. So, I believe the usage of `count` as static is good in this case.

plugins/externalwrench/src/ExternalWrench.cc
Yeshasvitvs: Time handling is updated https://github.com/robotology/gazebo-yarp-plugins/pull/418/commits/64199f6306c41ef36020454d52dfa14c72b24a8a

plugins/externalwrench/include/gazebo/ExternalWrench.hh
traversaro: And what about two  different externalwrench plugins in two different model that are spawned together:  the count would be shared also in that casel

plugins/externalwrench/include/gazebo/ExternalWrench.hh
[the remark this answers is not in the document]
traversaro: You obtain this behavior just by changing `std::unique_ptr<wrenchCommand> wrenchPtr{new wrenchCommand()};` to `wrenchCommand wrench;`? If that is the case, it is extremely strange and I do not know what's going  on. 

If instead you obtain  this because you changed `boost::shared_ptr<ExternalWrench> newWrench(new ExternalWrench);` to `ExternalWrench newWrench` and other changes related to `ExternalWrench`,  I guess what is going on is that there is something going on related to the use of `static count`, that is increased whenever a new copy of `ExternalWrench` is created, but not when a new copy of the shared_ptr is created. This is exactly the tricky behavior related to static variables (that are actually just global variables a bit disguised) that I  was mentioning in  https://github.com/robotology/gazebo-yarp-plugins/pull/418/files#r273442326 . 
I think the plugin will be much more maintainable if you just drop the `static count` and just  have a normal local `index` or `count` attribute that can be set with a `SetIndex` or `SetCount` method, that can contain all the code currently contained in the constructor. This `SetIndex`/`SetCount` method can be set directly in the https://github.com/robotology/gazebo-yarp-plugins/pull/418/files#diff-6f34d47dcd5558cbef158050b5563c76R175 , using some kind of  variable shared between `RPCServerThread` and  `ApplyExternalWrench` that is increase/decreased appropriately.

plugins/externalwrench/include/gazebo/ExternalWrench.hh
traversaro: While writing  https://github.com/robotology/gazebo-yarp-plugins/pull/418/files#r273692453 , I am a bit confused by  the `count` logic  even in the case of the single plugin. Let's say that three wrenches are added,  and they get the count `1`, `2` and `3` , where `1` and `3` have a duration of 100 seconds, and `2` instead has a duration of 1 seconds. After the the wrench 2 is deleted, the count counter is decremented to 2. If (let's say after 10 seconds) a new wrench is added, the `count` counter is increased again, and so the new wrench gets a `count` of 3, that is a index that is already used  by the wrench  `3`, and the visual representation will probably conflict!

plugins/externalwrench/include/gazebo/ExternalWrench.hh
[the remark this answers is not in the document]
Yeshasvitvs: This behavior is after changing `std::unique_ptr<wrenchCommand> wrenchPtr{new wrenchCommand()}; to wrenchCommand wrench;` and also `boost::shared_ptr< std::vector< boost::shared_ptr<ExternalWrench>>> wrenchesVectorPtr{new std::vector< boost::shared_ptr<ExternalWrench>>()};` to `std::vector<ExternalWrench> wrenchesVector`. Anyway, I do agree that `static count` variable will eventually cause problems when having multiple model having the plugin. So, I will have a variable `wrenchesIndex` in `ApplyExternalWrench` class and set the `wrenchIndex` variable in `ExternalWrench` class.

plugins/externalwrench/include/gazebo/ExternalWrench.hh
Yeshasvitvs: The conflict will happen for the visuals. So, it is better to not decrease the count of the wrenches after the duration is done. It is better to keep an increasing `wrenchCount` variable in `ApplyExternalWrench` and set the `wrenchIndex` variable of `ExternalWrench` class. This way each new wrench will have a different color without conflicts.

plugins/externalwrench/include/gazebo/ExternalWrench.hh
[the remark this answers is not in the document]
Yeshasvitvs: `static count` is removed and updated with new wrench count and index variables https://github.com/robotology/gazebo-yarp-plugins/pull/418/commits/ada82910ef25b7c3875faffd149d8802566f211a

Shared points are removed and simple data structures are used https://github.com/robotology/gazebo-yarp-plugins/pull/418/commits/b7d1d383c7b4ed478092b07945afe13e13960b4e

Yeshasvitvs:
@traversaro any update on this ?

traversaro:
>@traversaro any update on this ?

Sorry @Yeshasvitvs , I do not get notifications for new commits pushed in the PR, and even if I got them I  have no way to know if all the review comments has been addressed. From your comment I imagine that the PR is ready for a new review round, but for the future feel free to comment on the PR when the review comments has been addressed, so the reviewer can check the new code.

Yeshasvitvs:
All the previous reviews comments are address in the recent commits 😉 please check them and let me know if there are more changes needed. Thanks @traversaro

traversaro:
There are some comments below, but nothing blocking the merge. Do you want to cleanup the commits or otherwise I will squash the commits as part of the merge. 

Now the `ExternalWrench::setWrenchColor` function will take more and more time as simulation goes on, am I right. This is probably not desirable, especially if the `wrenchIndex` will grow a lot (something like ~10000 or similar).

Yeshasvitvs:
@traversaro Now the ExternalWrench::setWrenchColor function will take more and more time as simulation goes on

yes that is the way it is coded now

> This is probably not desirable, especially if the wrenchIndex will grow a lot (something like ~10000 or similar)

Yes but it is an extreme case. Can you think of any other way to handle the uniform distribution through `ApplyExternalWrench` class ??

traversaro:
> Yes but it is an extreme case. Can you think of any other way to handle the uniform distribution through ApplyExternalWrench class ??

~~~
std::mt19937 rand_gen(wrench_index);
std::uniform_real_distribution<> uni_dis(0.0, 1.0)
~~~

Yeshasvitvs:
> > Yes but it is an extreme case. Can you think of any other way to handle the uniform distribution through ApplyExternalWrench class ??
> 
> ```
> std::mt19937 rand_gen(wrench_index);
> std::uniform_real_distribution<> uni_dis(0.0, 1.0)
> ```

@traversaro updated color handling using wrench index as seed https://github.com/robotology/gazebo-yarp-plugins/pull/418/commits/c4da8fb6a480d77e2d785f9d30935c40e0b1cd0e

traversaro:
Ok, do you want to cleanup the commits? Otherwise I will squash the commits as part of the merge.

Yeshasvitvs:
@traversaro please take care of the commits while merging. Thanks

Issue #421, opened 2019-04-18
Cleanup unused CMake checks


traversaro:
In the CMake's build scripts, there are a lot of checks of `CMAKE_VERSION` ( https://github.com/robotology/gazebo-yarp-plugins/search?p=3&q=CMAKE_VERSION&type=&utf8=%E2%9C%93  ), there are not useful as now we require CMake >= 3.5 . It would be a good idea to remove them to cleanup the code.

MiladShafiee:
@traversaro I'm interested to do this but I don't have any idea about that! :wink:

traversaro:
Great, let's discuss this f2f when you want. For a preliminary reading, I suggest to read the YARP's contributing guide, see https://github.com/robotology/yarp/blob/master/.github/CONTRIBUTING.md#workflow .

traversaro:
Fixed by @MiladShafiee in https://github.com/robotology/gazebo-yarp-plugins/pull/423, thanks!

PR #423, opened 2019-04-18
Removing unused CMake checks for version less than 3.5 (#421)

MiladShafiee:
There was some unused checking for the minimum version required of CMake in CMakes build scripts, so we removed these additional checking as described in this issue  #421!


MiladShafiee:
> Great, thanks a lot.
> 
> As you removed the workaround for the linking of `gazebo_yarp_lib_common`, I think a further step in the simplification would be to remove the set of the `LIB_COMMON_NAME` and directly pass the `gazebo_yarp_lib_common` to the `LINKED_LIBRARIES` option of the `add_gazebo_yarp_plugin_target` CMake function. However, we can also merge as it is, as it is already an improvement.

@traversaro  Thank you! I've done the proposed modification!

CMakeLists.txt.user
```diff
@@ -0,0 +1,421 @@
...
+ </data>
+ <data>
+  <variable>ProjectExplorer.Project.Updater.FileVersion</variable>
+  <value type="int">20</value>
+ </data>
+ <data>
+  <variable>Version</variable>
+  <value type="int">20</value>
+ </data>
+</qtcreator>
```
traversaro: The `CMakeLists.txt.user` file is a file automatically generated by QtCreator, and it is not supposed to be committed to the repo.

MiladShafiee: Thank you. I removed this file.

traversaro:
Thanks!

PR #424, opened 2019-04-30
Add resetSimulationState and regenerate thrift-generated files

triccyx:
Reset position to initial pose, do not reset time.


traversaro:
What is the advantage of resetting the state but not resetting time?

triccyx:
This is a work-around in order to avoid a problem when resetting the time. If a yarp network wait is running
it will hang (not really hung but delayed) when the time is reset.

traversaro:
Given that we have already `resetSimulation` that resets time and state, and `resetSimulationTime` that resets only the time, I suggest to rename the method `resetSimulationState`, that just reset the state but not the time. This should make the interface more portable to other simulators and clarify better its behavior (for example, I guess the method also resets the **velocity** of the links and joints, not only its position.

plugins/clock/src/Clock.cc
```diff
@@ -202,7 +208,7 @@ namespace gazebo
             return physics->GetMaxStepSize();
         }
         return -1.0;
-    }
+    }
```
traversaro:
```suggestion
    }
```

traversaro:
Ok for me, modulo the small trailing whitespace comment. @diegoferigo any comment?

triccyx:
@traversaro  @diegoferigo  gentle ping 
Could you merge the pull request?

traversaro:
Thanks for the PR!

Issue #442, opened 2019-09-20
Fix  Travis on Devel branch


traversaro:

traversaro:
Now Travis is failing due to https://github.com/robotology/yarp/issues/2086 .

PR #443, opened 2019-09-30
gazebo_yarp_imu: add multipleanalogsensors interface

Nicogene:
This PR add the multiple analog interface to the `gazebo_yarp_imu` plugin.

Please review code.

(TESTED)

### DIFF
--- a/plugins/imu/include/gazebo/IMU.hh
+++ b/plugins/imu/include/gazebo/IMU.hh
@@ -10,6 +10,7 @@
 
 #include <gazebo/common/Plugin.hh>
 
+#include <yarp/dev/IMultipleWrapper.h>
 #include <yarp/dev/PolyDriver.h>
 
 #include <string>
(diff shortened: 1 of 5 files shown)


traversaro:
cc @nunoguedelha @prashanthr05 can you have a look? Thanks!

traversaro:
Travis is failing due to https://github.com/robotology/gazebo-yarp-plugins/issues/442 (TL;DR: we need to use YARP devel in the Travis script).

Nicogene:
This path has been successfully tested with these patches: 

https://github.com/robotology/icub-gazebo/pull/71 and https://github.com/robotology/icub-main

## iCubGui + inertial
ezgif com-video-to-gif(2)
## iCubGui + WBD
ezgif com-video-to-gif(3)

traversaro:
Thanks a lot @Nicogene ! Do you think it is possible to keep the plugin compatible with the old configuration data? As we have a few models around that assume the old format and we also keep models in a repo that does not have `master` and `devel` branches, and it would be nice to offer a smooth transition path. A possible strategy is to check for a parameter that enables the parsing according to the new format, and instead fallback to the old format, printing a deprecation. As soon as we can start working on gazebo-yarp-plugins 3.4 , we can drop this old compatibility logic, and just use the new logic that you prepared. Furthermore, can you update the changelog?

plugins/imu/include/gazebo/IMU.hh
```diff
@@ -45,11 +46,15 @@ namespace gazebo
...
         GazeboYarpIMU();
         virtual ~GazeboYarpIMU();
-        virtual void Load(sensors::SensorPtr _sensor, sdf::ElementPtr _sdf);
+        void Load(sensors::SensorPtr _sensor, sdf::ElementPtr _sdf) override;
 
     private:
         yarp::os::Property m_parameters; 
         yarp::dev::PolyDriver m_imuDriver;
+        yarp::dev::PolyDriver m_MASWrapper;
+        yarp::dev::PolyDriver m_AdditionalWrapper;
```
prashanthr05: Could you document why this additional wrapper is necessary?

traversaro:
For using the legacy behavior, we can just check if `device inertial` is present in the config file.

plugins/imu/include/yarp/dev/IMUDriver.h
```diff
@@ -55,24 +60,188 @@ class yarp::dev::GazeboYarpIMUDriver:
...
+    /* IThreeAxisGyroscopes methods */
+    /**
+     * Get the  number of three axis gyroscopes in the device
+     * @return 1
+     */
+    size_t getNrOfThreeAxisGyroscopes() const override;
+
+    /**
+     * Get the status of three axis gyroscope
+     * @param[in] sens_index sensor index (must be 0 in the case BoschIMU)
```
prashanthr05: `in the case BoschIMU` can be removed?

plugins/imu/include/yarp/dev/IMUDriver.h
```diff
@@ -55,24 +60,188 @@ class yarp::dev::GazeboYarpIMUDriver:
...
+    bool getOrientationSensorFrameName(size_t sens_index, std::string &frameName) const override;
+
+    /**
+     * Get orientation sensor measurements
+     * @param[in] sens_index sensor index (must be 0 in the case BoschIMU)
+     * @param[out] out RPY Euler angles in deg
+     * @param[out] timestamp timestamp of measurement
+     * @return true/false success/failure
+     */
+    bool getOrientationSensorMeasureAsRollPitchYaw(size_t sens_index, yarp::sig::Vector& rpy, double& timestamp) const override;
```
traversaro: I would just drop all the documentation of this methods. The methods are already documented in YARP, and copy pasting the docs just make them prone to be outdated if the YARP docs are updated.

prashanthr05:
I think the documentation could include how to write a sample configuration file accounting for the new features and the usual table for specifying the required parameters must be updated.

Otherwise, the code lgtm!

Nicogene:
Changes requested addressed, I updated the documentation and I add a fallback open in case you have a old version of the ini file

plugins/imu/include/yarp/dev/IMUDriver.h
```diff
@@ -55,24 +60,80 @@ class yarp::dev::GazeboYarpIMUDriver:
...
+    yarp::dev::MAS_status genericGetStatus(size_t sens_index) const;
+    bool genericGetSensorName(size_t sens_index, std::string &name) const;
+    bool genericGetFrameName(size_t sens_index, std::string &frameName) const;
+    bool genericGetMeasure(size_t sens_index, yarp::sig::Vector& out, double& timestamp, size_t startIdx) const;
+
     yarp::sig::Vector m_imuData; //buffer for imu data
     yarp::os::Stamp m_lastTimestamp; //buffer for last timestamp data
-    yarp::os::Semaphore m_dataMutex; //mutex for accessing the data
+    mutable yarp::os::Semaphore m_dataMutex; //mutex for accessing the data
+    std::string m_sensorName{"sensor_imu_gazebo"};
```
traversaro: Gazebo has a Sensor::Name method, I think we can use that one for both the sensor and the frame, at least for now. We can read it in the `Load` and then pass it to the device via the yarp parameters.

Nicogene: Done :+1:

traversaro: Sorry, I think it makes more sense (for eventual consistency with the real robot) to use the `::Name` method, not the `ScopedName` that will contains also the model name and the world name.

plugins/imu/src/IMU.cc
```diff
@@ -57,18 +75,92 @@ void GazeboYarpIMU::Load(sensors::SensorPtr _sensor, sdf::ElementPtr _sdf)
...
+        legacy_behaviour = true;
+    }
+
+    if (legacy_behaviour) {
+        m_parameters.put(YarpIMUScopedName.c_str(), m_scopedSensorName.c_str());
+        m_parameters.put("sensor_name",_sensor->Name());
+        if (!m_imuDriver.open(m_parameters)) {
+            yError() << "GazeboYarpIMU Plugin Load failed: error in opening yarp driver";
+        }
+        return;
```
prashanthr05: In this case, I assume we are inducing a wrong behavior, by not attaching itself to the analog server?

Nicogene: Actually inside this if statement there is the code as it was before, then if you have the old ini file it will open as `device` the `ServerInertial` and as `subdevice` the `GazeboYarpIMU`. No attach is required in this case as before.

plugins/imu/include/yarp/dev/IMUDriver.h
Nicogene: I used the `_sensor->Name()` check line 145 of IMU.cc

plugins/imu/src/IMU.cc
prashanthr05: Ah okay..! Understood, since we are dealing with `ServerInertial`. Thanks @Nicogene !

traversaro:
A few Travis build fails due to https://github.com/robotology/yarp/issues/2086 ,  but the rest is green. @prashanthr05 if you can approve, then we can merge.

plugins/imu/include/yarp/dev/IMUDriver.h
traversaro: You are right, I was confused by the old code that used `ScopedName` instead, thanks!

PR #457, opened 2019-12-02
Add gazebo_yarp_contactforcetorques

traversaro:
Add gazebo plugin simulating the estimation of contact force torques

This plugin does not simulate any physical sensor, but it simulates the output of the estimators of external contact force torques,
similarly to how the gazebo_yarp_basestate plugin simulates the output of floating base position estimators.

On real robot, the contact forcetorques are estimated by a dedicated software based on the available sensor measurements (such as motor currents, six axis force-torque sensors measures and distribute tactile sensor measurements) and/or a-priori assumptions on the location of the contact (for example on industrial manipulators the external contact is usually assume to be only possible on the end-effector). For example, on the iCub robot the [`wholebodydynamics` software](https://github.com/robotology/whole-body-estimators) based on the [`iDynTree::ExtWrenchesAndJointTorquesEstimator` class](https://robotology.github.io/idyntree/master/classiDynTree_1_1ExtWrenchesAndJointTorquesEstimator.html) is responsible for such an estimation. In a simulation, it is possible to obtain directly this information for the simulator state and inputs, if it is necessary to have a "perfect" estimate.

For more info, see the docs in the doxygen block of the GazeboYarpContactForceTorques, and the example in ` tutorial/model/contactforcetorques`.

### DIFF
--- a/plugins/CMakeLists.txt
+++ b/plugins/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_subdirectory(camera)
 add_subdirectory(clock)
+add_subdirectory(contactforcetorques)
 add_subdirectory(controlboard)
 
 # Disable externalwrench plugin if Gazebo version is < 8.0
--- a/plugins/contactforcetorques/CMakeLists.txt
+++ b/plugins/contactforcetorques/CMakeLists.txt
@@ -1,0 +1,16 @@
+# Copyright (C) 2019 Fondazione Istituto Italiano di Tecnologia
+# Authors: see AUTHORS file.
+# CopyPolicy: Released under the terms of the LGPLv2.1 or later, see LGPL.TXT
+
+include(AddGazeboYarpPluginTarget)
+
+add_gazebo_yarp_plugin_target(LIBRARY_NAME contactforcetorques
+                              INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR}
+                              INCLUDE_DIRS include/gazebo 
+                                           include/yarp/dev
+                              SYSTEM_INCLUDE_DIRS ${GAZEBO_YARP_COMMON_HEADERS} ${Boost_INCLUDE_DIRS} ${GAZEBO_INCLUDE_DIRS} ${SDFORMAT_INCLUDE_DIRS} ${PROTOBUF_INCLUDE_DIRS}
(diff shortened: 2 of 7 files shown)


traversaro:
The macOS failure is a unrelated to the PR, due to the new protobuf homebrew bottle.

Known issues: 
* Only a limited combinations of additional frames is currently supported, see https://github.com/robotology/gazebo-yarp-plugins/pull/457/files#diff-07f2c66ba9e962f4a48d69055a97a8cfR299 
* You need to also add a contact sensor to all the collisions attached to the link of which you want to measure the contact force, see for example https://github.com/robotology/gazebo-yarp-plugins/pull/457/files#diff-2e8dbd17a8bb2f9b9ee0316d2e3002ceR46 .

traversaro:
Further work is necessary, so this will not go in the 3.3 release.

GiulioRomualdi:
Hi @traversaro, I've just noticed this PR. Let me know if it is ready to be reviewed.

traversaro:
At the moment to get this PR to merge we need either to finish the implementation (see https://github.com/robotology/gazebo-yarp-plugins/pull/457/files#diff-07f2c66ba9e962f4a48d69055a97a8cfR299) or modify the docs to match the current status. I do not plan to work on it for the near future, so if anyone wants to work on it I would be happy to  provide support. 

To clarify, note that the main objective of this plugin is to provide the external contact force information in the same format used by `wholebodydynamcis`, but **without the need of actually running `wholebodydynamics`** . This is useful for: 
* Avoid the need to launch an external `yarprobotinterface` just to run  `wholebodydynamics`, simplifying the setup of simulations 
* Permit to run in simulation software that requires external contact force estimates (such as https://github.com/robotology/whole-body-controllers or https://github.com/robotology/walking-controllers) on models that are not equipped with internal six-axis FT sensors, that are the sensors used by wholebodydynamics to provide its estimates. 

fyi @robotology/iit-dynamic-interaction-control

prashanthr05:
I tested this PR in its current status with the modifications to iCubGazeboV2_5 in https://github.com/prashanthr05/icub-models/commit/cb6496c8791f47730ef04eec0ec66f740bfd5c13

It seems to be working as expected.

prashanthr05:
I tested this with WalkingModule. 
- At higher rates of this device, the robot keeps falling after two steps.
- At a nominal rate of 100 Hz, the Walking module fails to read the wrenches from the ports and closes.

Instead, WalkingModule does not fail/stop with wholebodydynamics device running.

traversaro:
@S-Dafarra @GiulioRomualdi @prashanthr05 I am going to close this for now, if anyone want to work again on it feel free to resume from this PR, thanks!

PR #461, opened 2020-01-27
Add joinType option to linkattacher

lrapetti:
This PR add the possibility to use `linkattacher` plugin creating type of joint different from `fixed`. In particular, the fixed joint will remain the default type, but any other existing jointType can be used by defining `jointType` in the configuration as explained in https://github.com/lrapetti/gazebo-yarp-plugins/tree/update-linkattacher/plugins/linkattacher#usage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
 
 The format of this document is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Added
+- Add the possibility to create different types of joints with the `linkattacher` plugin (https://github.com/robotology/gazebo-yarp-plugins/pull/461).
+
 ## [3.3.0] - 2019-12-13
 
 ### Added
--- a/plugins/linkattacher/src/linkattacher.cc
+++ b/plugins/linkattacher/src/linkattacher.cc
@@ -53,6 +53,21 @@ void LinkAttacher::Load(physics::ModelPtr _model, sdf::ElementPtr _sdf)
       return;
   }
 
+  std::string jointType;
+  if(m_parameters.check("jointType"))
+  {
+      jointType = m_parameters.find("jointType").asString();
+  }
+  else
+  {
+      jointType="fixed";
+  }
+
+  if(!m_la_server.setJointType(jointType))
+  {
+      return;
+  }
+  
   std::string portname;
   if(m_parameters.check("name"))
   {
@@ -72,5 +87,4 @@ void LinkAttacher::Load(physics::ModelPtr _model, sdf::ElementPtr _sdf)
   }
 
   m_la_server.yarp().attachAsServer(*m_rpcport);
-
 }
--- a/plugins/linkattacher/src/linkattacherserverimpl.cpp
+++ b/plugins/linkattacher/src/linkattacherserverimpl.cpp
@@ -6,6 +6,9 @@
 
 #include <linkattacherserverimpl.h>
 
+#include <algorithm>
+#include <vector>
+
 using namespace gazebo;
 using namespace std;
 using namespace yarp::os;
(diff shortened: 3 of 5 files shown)


lrapetti:
The only problem with this implementation involves the fact that when the `jointType` does not exists, calling [`_world->Physics()->CreateJoint(jointType,parent_model)`](https://github.com/robotology/gazebo-yarp-plugins/pull/461/files#diff-7e1089c30889bd5585eeb97aae62e840R98) raises an exception error with no info. 
```
libc++abi.dylib: terminating with uncaught exception of type gazebo::common::Exception
gzclient(28783,0x1160b85c0) malloc: Incorrect checksum for freed object 0x7fcd835c39d0: probably modified after being freed.
Corrupt value: 0x746c7561666564
gzclient(28783,0x1160b85c0) malloc: *** set a breakpoint in malloc_error_break to debug
```
I don't know if this should be somehow handled.

plugins/linkattacher/src/linkattacher.cc
```diff
@@ -53,6 +53,18 @@ void LinkAttacher::Load(physics::ModelPtr _model, sdf::ElementPtr _sdf)
...
+  if(m_parameters.check("jointType"))
+  {
+      jointType = m_parameters.find("jointType").asString();
+  }
+  else
+  {
+      jointType="fixed";
+  }
+  m_la_server.setJointType(jointType);
+  yInfo() << LogPrefix << "joint type: " << jointType;
```
traversaro: Please remove, thanks!

plugins/linkattacher/src/linkattacherserverimpl.cpp
```diff
@@ -92,34 +92,34 @@ bool LinkAttacherServerImpl::attachUnscoped(const string& parent_model_name, con
...
     //This is joint creation
-    gazebo::physics::JointPtr fixed_joint;
+    gazebo::physics::JointPtr joint;
 
     #if GAZEBO_MAJOR_VERSION >= 8
-    fixed_joint = _world->Physics()->CreateJoint("fixed",parent_model);
+    joint = _world->Physics()->CreateJoint(jointType,parent_model);
     #else
-    fixed_joint = _world->GetPhysicsEngine()->CreateJoint("fixed",parent_model);
+    joint = _world->GetPhysicsEngine()->CreateJoint("fixed",parent_model);
```
traversaro: Any reason why `fixed` is not changed to `jointType` here?

plugins/linkattacher/src/linkattacher.cc
lrapetti: done https://github.com/robotology/gazebo-yarp-plugins/pull/461/commits/c2707e4f9eaf9c1e62023d037e944bce673bd357

plugins/linkattacher/src/linkattacherserverimpl.cpp
lrapetti: leftover, fixed in https://github.com/robotology/gazebo-yarp-plugins/pull/461/commits/acd167ae506eae4492220c4fa4db6cdbcb718f79

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 ### Added
 - Add the usage of `MultipleAnalogInterfaces`(https://www.yarp.it/group__dev__iface__multiple__analog.html)
   in the `gazebo_yarp_imu`.
+- Add the possibility to create different types of joints with the `linkattacher` plugin (https://github.com/robotology/gazebo-yarp-plugins/pull/461).
```
lrapetti: wrong location, should move it among unreleased

lrapetti: done https://github.com/robotology/gazebo-yarp-plugins/pull/461/commits/6917e052926ff0020873cfba94c756248d0c4b84

traversaro:
> 
> 
> The only problem with this implementation involves the fact that when the `jointType` does not exists, calling [`_world->Physics()->CreateJoint(jointType,parent_model)`](https://github.com/robotology/gazebo-yarp-plugins/pull/461/files#diff-7e1089c30889bd5585eeb97aae62e840R98) raises an exception error with no info.
> 
> ```
> libc++abi.dylib: terminating with uncaught exception of type gazebo::common::Exception
> gzclient(28783,0x1160b85c0) malloc: Incorrect checksum for freed object 0x7fcd835c39d0: probably modified after being freed.
> Corrupt value: 0x746c7561666564
> gzclient(28783,0x1160b85c0) malloc: *** set a breakpoint in malloc_error_break to debug
> ```
> 
> I don't know if this should be somehow handled.

I would list explicitly the supported joint types, and fail early with a clear error message if one of the non-supported types is passed.

lrapetti:
> I would list explicitly the supported joint types, and fail early with a clear error message if one of the non-supported types is passed.

I think this can be a solution. However, I am not sure what documentation I should refer to list the available joints in Gazebo, since as far as I understood, it depends on the Physics Engine .

A possibility, is that I might refer to [`SDF` documentation](http://sdformat.org/spec?ver=1.6&elem=joint#joint_type) regarding joint types.

traversaro:
> A possibility, is that I might refer to [`SDF` documentation](http://sdformat.org/spec?ver=1.6&elem=joint#joint_type) regarding joint types.

Yes, let's refer to that and only allow those values. In the remote case in which a new joint type will be added to SDF, we will update the plugin.

lrapetti:
> > A possibility, is that I might refer to [`SDF` documentation](http://sdformat.org/spec?ver=1.6&elem=joint#joint_type) regarding joint types.
> 
> Yes, let's refer to that and only allow those values. In the remote case in which a new joint type will be added to SDF, we will update the plugin.

I have add the check with https://github.com/robotology/gazebo-yarp-plugins/pull/461/commits/99485d1f7b447bf58270943ac6cec11367d44b8b, and I have also tested all the available joint types noting that `gearbox` joint type is currently not working (gazebo crashes as if the joint type doesn't exist). I have updated the documentation accordingly with https://github.com/robotology/gazebo-yarp-plugins/pull/461/commits/00b271123ca0870c621d88a442d1174551068051.

lrapetti:
I have cleaned the commit history, waiting for the checks to merge.

traversaro:
> 
> 
> I have cleaned the commit history, waiting for the checks to merge.

CI is fine, merged. Thanks for the contribution. We do not plan for any feature release in the near future, if you need this change out for 2020.02 feel free to ping me (see https://github.com/robotology/robotology-superbuild/issues/331).

Issue #460, opened 2020-01-16
gazebo_yarp_controlboard: Changing the interaction mode to stiff of a joint that is already in stiff mode resets the PID


MiladShafiee:
I want to change the control mode of the some joints during walking from stiff to compliant  (or vice versa).
For debugging, I try to re-set at each 100 hundred cycle the mode of the joints to stiff mode in `updateModule()` function in the walking controller by following function:

```
    if(!m_InteractionInterface->setInteractionModes(m_JointModeStiffVectorDefult.data()))
    {
         yError() << "[RobotInterface::setIntractionModeToStiffMode] Error while setting the stiff interaction modes of the joints";
        return false;
    }
```
that `m_JointModeStiffVectorDefult` is:
```
        for (unsigned int i=0;i<m_actuatedDOFs;i++)
        {
             m_JointModeStiffVectorDefult.at(i)=yarp::dev::InteractionModeEnum::VOCAB_IM_STIFF;
        }
```

I run the walkingModule and the behavior is as following:

![ezgif com-video-to-gif (16)](https://user-images.githubusercontent.com/31568170/72532283-6a0f3f00-3873-11ea-8d8a-29d45f7b6117.gif)

traversaro:
Hi @MiladShafiee , 

>  each 100 hundred cycle

Each 100 hundred cycle of what? 

> I run the walkingModule and this is the behavior!

So, you basically already have the joints in stiff mode, and you get that strange behaviour when you try to "switch again" in stiff mode? Probably there is some wrong reset that is happening when the  stiff mode is set.

traversaro:
The problem is  in https://github.com/robotology/gazebo-yarp-plugins/blob/3fa2c45a4b8a63ef15c524462124b505f29b3ed9/plugins/controlboard/src/ControlBoardDriverInteractionMode.cpp#L43 . The PID are reset (including its integral term, that is reset to 0) and the setpoint is reset to the current position when changing the control mode, but they should change only if the specified interaction mode is different from the current one, otherwise the function should return without any reset. A Pull Request to address this is welcome.

traversaro:
Note that it may be possible that such behavior is present also in physical robots. Furthermore, changing the interaction mode may be an expensive operation in physical robots (the low-level controlboards need to change their  control law, etc etc), so in general it may not be ideal to set the interaction  mode every one second (you can add a check in your code to only set the interaction mode to stiff if it is not stiff at the moment).

[...]

PR #464, opened 2020-01-28
solve the problem of change of the interaction mode if you are already on that mode

MiladShafiee:
Here we solved the problem of asking to change the interaction mode, If we are already on that mode. For example if we are in stiff mode and we ask again to set to stiff mode, we shouldn't reset again to stiff mode because it resets the PID.
The related Issue #460

### DIFF
--- a/plugins/controlboard/src/ControlBoardDriverInteractionMode.cpp
+++ b/plugins/controlboard/src/ControlBoardDriverInteractionMode.cpp
@@ -40,10 +40,10 @@ bool GazeboYarpControlBoardDriver::getInteractionModes(yarp::dev::InteractionMod
 
 bool GazeboYarpControlBoardDriver::changeInteractionMode(int j, yarp::dev::InteractionModeEnum mode)
 {
-	// The joint is already in the specified interaction mode, return
+    // The joint is already in the specified interaction mode, return
     // See https://github.com/robotology/gazebo-yarp-plugins/issues/460 for more details
     if (m_interactionMode[j] == mode) {
-    return true;
+        return true;
     }
 
     resetAllPidsForJointAtIndex(j);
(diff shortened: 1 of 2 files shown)


plugins/controlboard/src/ControlBoardDriverInteractionMode.cpp
```diff
@@ -40,6 +40,12 @@ bool GazeboYarpControlBoardDriver::getInteractionModes(yarp::dev::InteractionMod
 
 bool GazeboYarpControlBoardDriver::changeInteractionMode(int j, yarp::dev::InteractionModeEnum mode)
 {
+	// The joint is already in the specified interaction mode, return
```
traversaro: Please avoid to mix tabs and spaces, and indent correctly, thanks!
```suggestion
    // The joint is already in the specified interaction mode, return
```

plugins/controlboard/src/ControlBoardDriverInteractionMode.cpp
```diff
@@ -40,6 +40,12 @@ bool GazeboYarpControlBoardDriver::getInteractionModes(yarp::dev::InteractionMod
...
+    // See https://github.com/robotology/gazebo-yarp-plugins/issues/460 for more details
+    if (m_interactionMode[j] == mode) {
+    return true;
```
traversaro: Please indent correctly the code, thanks!
```suggestion
        return true;
```

plugins/controlboard/src/ControlBoardDriverInteractionMode.cpp
MiladShafiee: I modified. Thanks

plugins/controlboard/src/ControlBoardDriverInteractionMode.cpp
MiladShafiee: I modified. Thanks

MiladShafiee:
@traversaro Should I do something else or is there any problem in the modifications?

traversaro:
> @traversaro Should I do something else or is there any problem in the modifications?

Sorry, I forgot to reply to you, thanks, the PR is ready, I just restarted the CI to verify that is still compiling, then we can merge. 
If I fail to reply, feel free to ping me in the future.

MiladShafiee:
> Sorry, I forgot to reply to you, thanks, the PR is ready, I just restarted the CI to verify that is still compiling, then we can merge.
> If I fail to reply, feel free to ping me in the future

Thank you.

traversaro:
Thanks!

Issue #465, opened 2020-03-04
iCub: iPositionControl: relativeMove execute correctly but not move the joints


torstenfol:
Hey,
I am working with the iCub in Chemnitz. While working with the iCub in the gazebo simulator a come across a problem with the motor control:
I have a problem in using the relativeMove() methods of the IPositionControl interface.
While with the positionMove the joints can be controlled correctly, the joints did not move with the use of relativeMove. This problem does not occur with the iCub-simulator, there the relativeMove are executed correctly.

used Software:
Gazebo V11
Yarp V3.3.101
Yarp-Gazebo-Plugins V3.3.0

Best,
Torsten Follak

traversaro:
Hi @torstenfol, thanks for reporting the issue. Can you check if the modification in https://github.com/robotology/gazebo-yarp-plugins/pull/466 fixes the problem for you? Thanks!

torstenfol:
@traversaro thank you, I will test it and report the result.

torstenfol:
It solved the problem and it works now correctly! Thank you!

traversaro:
Fixed in gazebo-yarp-plugins v3.3.1, see : https://github.com/robotology/gazebo-yarp-plugins/releases/tag/v3.3.1 .

PR #466, opened 2020-03-04
Fix relativeMove implementation

traversaro:
`GazeboYarpControlBoardDriver::relativeMove` was not correctly implemented until now.
See https://www.yarp.it/classyarp_1_1dev_1_1IPositionControl.html#a32763b9f36c8751b589b23ff79915e0b for the documentation on how the `yarp::dev::IPositionControl::relativeMove` method should be implemented. 
 
Fix https://github.com/robotology/gazebo-yarp-plugins/issues/465 .


traversaro:
Thanks @lrapetti !

Issue #320, opened 2017-09-25
delete virtual objects in Gazebo through worldInterface plugin


pattacini:
@valeriosperati commented on [Tue Jul 04 2017](https://github.com/robotology/QA/issues/241)

Hi, 
I successfully created a cylinder in gazebo using the worldInterface plugin.  
I did it connecting my c++ code to  the port "/world_input_port", and then writing those commands, as suggested in [#201](https://github.com/robotology/QA/issues/201) by @pattacini :

```c++
yarp::os::Bottle cmd, rep;

cmd.addString("makeCylinder");
cmd.addDouble(0.05); //radius [mt]
cmd.addDouble(0.1);  //lenght x  [mt]
cmd.addDouble(1.4);  //pos x  [mt]
cmd.addDouble(0.8);  //pos y	    
cmd.addDouble(0.3);  //pos z
cmd.addDouble(0.1);  //pose.roll  [rad] 
cmd.addDouble(0.1);  //pose.pitch [rad]	    
cmd.addDouble(0.1);  //pose.yaw   [rad]
cmd.addInt(0);     //color.r 
cmd.addInt(255);   //color.g 
cmd.addInt(0);     //color.b 

portToGazeboWorld.write(cmd, rep);
```

Now, the problem is that I cannot delete the object. Looking at [**Thrift**](https://github.com/robotology/gazebo-yarp-plugins/blob/master/thrift/worldinterface/worldinterface_rpc.thrift)  interface I wrote this command at the end of my program:

```c++
yarp::os::Bottle cmd, rep;
cmd.addString("deleteAll");           
portToGazeboWorld.write(cmd, rep);
std::string response=rep.get(0).asString();  
cout<<"Operation: "<<response<<"\n\n";
```
I guess the syntax is correctly received by gazebo, as the response is 
`Operation: ok`

Nevertheless, the cylinder remains in gazebo. What am I missing?

PS: I noted in the gazebo terminal a message appearing as soon as the object is created
`[WARNING]Internal error during object creation. Unimplemented feature in gazebo 7.`
 

---

@valeriosperati commented on [Tue Jul 04 2017](https://github.com/robotology/QA/issues/241#issuecomment-312914186)

I realized right now it happens with cylinder object only. Spheres and boxes behave as expected.
I get no name back from a cylinder, while I get names from sphere and box. 

---

@pattacini commented on [Mon Sep 25 2017](https://github.com/robotology/QA/issues/241#issuecomment-331909757)

[...]

PR #467, opened 2020-03-04
Worldinterface plugin: Implementation of loadModel; fixed makeCylinder return no name

torsten93:
Since I wanted to create objects in gazebo from python, I implemented the loadModel method in the worldinterface to spawn complex models.

Also the problem with the makeCylinder (#320) method is fixed by adjusted it to the makeBox method behavior.

Best,
Torsten


traversaro:
Hi @torstenfol, 
   thanks for the contribution. 

A small preliminary comment: as this is a new feature, could you open the new PR against the `devel` branch? I plan to release a feature version from the devel branch before may, so I think it is a good idea to keep this  new development in devel, so that it will be part of the new 3.4 release, thanks!

torstenfol:
Hi @traversaro,
I changed the PR base to devel branch. Is this okay or is a new PR needed?

traversaro:
> 
> 
> Hi @traversaro,
> I changed the PR base to devel branch. Is this okay or is a new PR needed?

Thanks, this  is ok!

plugins/worldinterface/src/worldproxy.cpp
```diff
@@ -181,7 +190,7 @@ std::string WorldProxy::makeSphere(const double radius, const GazeboYarpPlugins:
     if (collision_enable) {tmp->SetCollideMode("all");}
     else {tmp->SetCollideMode("none");}
   }
-  //HERE TMP IS ALWAYS NULL, so the following insertion is not valid. I DON'T KNOW TO FIX IT.
+  //HERE TMP IS ALWAYS NULL, so the following insertion is not valid. I DON'T KNOW TO FIX IT. -> fixed by waiting for insert
```
traversaro, line 193: Thanks, if the problem has been fixed, can you remove the old comment? Thanks!

traversaro:
Hi @torstenfol, sorry for the late review, if you are still interested in getting this merged, I have a few observations. 

* Instead of taking an existing function and changing its signature, could you create a function with a new name (for example `loadModelFromFileWithPose` and just define the old function by calling the new function)? In this way, we avoid breaking the code of who is using the old function. 
* Thanks for sorting out the problem of the null pointer due to the missing wait. However, I think that the current logic of checking the number of models is a bit brittle: what happens if in the same time that you are trying to insert the model someone removed a model? In that case the model count changed, but it is not related to the model that you spawned. I think directly checking if the model has been spawned is less error prone. 
* To ensure that you are actually waiting for real time and not against simulated time, I would wait on the system clock, see http://www.yarp.it/classyarp_1_1os_1_1SystemClock.html#ad5fba465106ea82a1ee037c72727cced .
* It would be ideal if the timeout could be specified, but this can be done in a follow up PR, I think we can also merge this PR with the hardcoded timeout. 
* If the timeout passed and the model has not been spawned, I think we should not go forward, ma return the function with some kind of value that signal an error to the caller.

In any case, thanks a lot for the contribution! 

FYI @randaz81

torstenfol:
Hi @traversaro thank you for the review. of course I am still interested in merging. I will include the proposed changes and update the PR.

torsten93:
Hi @traversaro, I implemented the requested changes, including the timeout as argument for the loadModel function. The same could make sense for the make...-functions, because there is the same check.
Since this would change the signature of all this function I hardcode there for now, but if it would be okay to have it for example as an optional argument I could add this here. Otherwise it could stay for now hardcoded and it would be at any time trivial to make a new function with this as an extra argument.

traversaro:
Thanks @torsten93 , the PR seems fine now.

traversaro:
I would leave the PR open for a few days to see if someone else wants to review it, and then we will merge it during next week. In the meanwhile, can you add a line describing your changes in https://github.com/robotology/gazebo-yarp-plugins/blob/devel/CHANGELOG.md#unreleased ? Thanks!

torsten93:
Okay, thanks. I will add the necessary lines in the next days.

torsten93:
@traversaro the entries in the changelog are done

traversaro:
Thanks @torsten93 !

PR #469, opened 2020-03-05
Added finger coupling management for iCub new Mk3 hand.

ale-git:
I've added the coupling class of the new iCub hand under developmente at present.
Since the coupling between the second and first phalanges is not linear, and thus not easily invertible, it makes use of an interpolated look up table that will be also useful in firmware implementation in control boards.

### DIFF
--- a/plugins/controlboard/include/yarp/dev/ControlBoardDriverCoupling.h
+++ b/plugins/controlboard/include/yarp/dev/ControlBoardDriverCoupling.h
@@ -164,4 +164,27 @@ public:
     yarp::sig::Vector decoupleRefTrq (yarp::sig::Vector& trq_ref);
 };
 
+class HandMk3CouplingHandler : public BaseCouplingHandler
+{
+
+public:
+    HandMk3CouplingHandler (gazebo::physics::Model* model, yarp::sig::VectorOf<int> coupled_joints, std::vector<std::string> coupled_joint_names);
+
+public:
+    bool decouplePos (yarp::sig::Vector& current_pos);
+    bool decoupleVel (yarp::sig::Vector& current_vel);
+    bool decoupleAcc (yarp::sig::Vector& current_acc);
+    bool decoupleTrq (yarp::sig::Vector& current_trq);
+
+    yarp::sig::Vector decoupleRefPos (yarp::sig::Vector& pos_ref);
+    yarp::sig::Vector decoupleRefVel (yarp::sig::Vector& vel_ref);
+    yarp::sig::Vector decoupleRefTrq (yarp::sig::Vector& trq_ref);
+    
+protected:
+    double decouple (double q2, double lut[]);
+
+    double thumb_lut[4096];
+    double index_lut[4096];
+};
+
 #endif //GAZEBOYARP_COUPLING_H
--- a/plugins/controlboard/include/yarp/dev/ControlBoardDriverCoupling.h
+++ b/plugins/controlboard/include/yarp/dev/ControlBoardDriverCoupling.h
@@ -169,6 +169,7 @@ class HandMk3CouplingHandler : public BaseCouplingHandler
 
 public:
     HandMk3CouplingHandler (gazebo::physics::Model* model, yarp::sig::VectorOf<int> coupled_joints, std::vector<std::string> coupled_joint_names);
+    virtual ~HandMk3CouplingHandler();
 
 public:
     bool decouplePos (yarp::sig::Vector& current_pos);
@@ -183,8 +184,10 @@ public:
 protected:
     double decouple (double q2, double lut[]);
 
-    double thumb_lut[4096];
-    double index_lut[4096];
+    const int LUTSIZE;
+
+    double *thumb_lut;
+    double *index_lut;
 };
 
 #endif //GAZEBOYARP_COUPLING_H
--- a/plugins/controlboard/include/yarp/dev/ControlBoardDriverCoupling.h
+++ b/plugins/controlboard/include/yarp/dev/ControlBoardDriverCoupling.h
@@ -169,7 +169,6 @@ class HandMk3CouplingHandler : public BaseCouplingHandler
 
 public:
     HandMk3CouplingHandler (gazebo::physics::Model* model, yarp::sig::VectorOf<int> coupled_joints, std::vector<std::string> coupled_joint_names);
-    virtual ~HandMk3CouplingHandler();
 
 public:
     bool decouplePos (yarp::sig::Vector& current_pos);
@@ -182,12 +181,12 @@ public:
     yarp::sig::Vector decoupleRefTrq (yarp::sig::Vector& trq_ref);
     
 protected:
-    double decouple (double q2, double lut[]);
+    double decouple (double q2, std::vector<double>& lut);
 
     const int LUTSIZE;
 
-    double *thumb_lut;
-    double *index_lut;
+    std::vector<double> thumb_lut;
+    std::vector<double> index_lut;
 };
 
 #endif //GAZEBOYARP_COUPLING_H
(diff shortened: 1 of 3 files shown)


plugins/controlboard/src/ControlBoardDriverCoupling.cpp
```diff
@@ -564,3 +565,231 @@ yarp::sig::Vector CerHandCouplingHandler::decoupleRefTrq (yarp::sig::Vector& trq
...
+        for (int n = 0; n < 4096; ++n)
+        {
+            if (num[n] > 0.0)
+            {
+                thumb_lut[n] /= num[n];
+            }
+        }
+    }
+    
+    //for (int n=0; n<1500; ++n) yDebug() << thumb_lut[n];
```
traversaro: I guess this can be removed?
```suggestion
```

plugins/controlboard/include/yarp/dev/ControlBoardDriverCoupling.h
```diff
@@ -164,4 +164,27 @@ class CerHandCouplingHandler : public BaseCouplingHandler
...
+    bool decoupleTrq (yarp::sig::Vector& current_trq);
+
+    yarp::sig::Vector decoupleRefPos (yarp::sig::Vector& pos_ref);
+    yarp::sig::Vector decoupleRefVel (yarp::sig::Vector& vel_ref);
+    yarp::sig::Vector decoupleRefTrq (yarp::sig::Vector& trq_ref);
+    
+protected:
+    double decouple (double q2, double lut[]);
+
+    double thumb_lut[4096];
```
traversaro: Instead of duplicating the value 4096 around, could be set the 4096 as a constexpr integer attribute and reuse it in the class? To be honest I am not sure if this works to set the size of a C-style array, but it should be easy to try.

ale-git: Done.

ale-git: It didn't work so I made the look up tables dynamically allocated.

plugins/controlboard/src/ControlBoardDriverCoupling.cpp
```diff
@@ -564,3 +565,239 @@ yarp::sig::Vector CerHandCouplingHandler::decoupleRefTrq (yarp::sig::Vector& trq
...
+HandMk3CouplingHandler::HandMk3CouplingHandler(gazebo::physics::Model* model, yarp::sig::VectorOf<int> coupled_joints, std::vector<std::string> coupled_joint_names)
+: BaseCouplingHandler(model, coupled_joints,coupled_joint_names), LUTSIZE(4096)
+{
+    const double RAD2DEG = 180.0/atan2(0.0,-1.0);
+    const double DEG2RAD = 1.0/RAD2DEG;
+    
+    m_couplingSize = 11;
+    
+    thumb_lut = new double[LUTSIZE];
+    index_lut = new double[LUTSIZE];
```
traversaro: Unfortunately manually managing this memory means that we would need to also define a set of custom copy constructors, copy operator, move constructors and move operators. I guess that perhaps it could be more simple to just use a `std::vector` and observe the [rule of zero](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-zero).

plugins/controlboard/src/ControlBoardDriverCoupling.cpp
```diff
@@ -564,3 +565,239 @@ yarp::sig::Vector CerHandCouplingHandler::decoupleRefTrq (yarp::sig::Vector& trq
...
+    
+    double num[LUTSIZE];
```
traversaro: Unfortunately Variable Length Arrays (VLA) are not part of the C++ standard (see https://stackoverflow.com/questions/1887097/why-arent-variable-length-arrays-part-of-the-c-standard), so to avoid problems in the future it is better to avoid their use. As this function is not called at every simulation step, I think it is ok to just use a `std::vector` here as well: 
```suggestion
    std::vector<double> num(LUTSIZE);
```

plugins/controlboard/src/ControlBoardDriverCoupling.cpp
ale-git: @traversaro done.

traversaro: Thanks!

ale-git:
Ciao @traversaro, can this be merged now?

traversaro:
> 
> 
> Ciao @traversaro, can this be merged now?

Sorry, I had a leftover review comment that was not sent, yes I think we can merge it now.

traversaro:
Ah, given the recent work on coupling controlboard perhaps @xEnVrE may be also interested in reviewing this. 

cc @randaz81

randaz81:
LGTM 👍 
Just for additional clarity, can you add to the description of this PR the following infos (if available):
1) a link to the relevant hand documentation (e.g. its geometry...)
2) a link to the sdf model of the robot which uses the new hand mk3.

xEnVrE:
I agree with @randaz81. Also, some of the implemented equations are complex and if, in the future, someone want to make a comparison between the implemented code and the documentation, it would be great if the equations are available somewhere.

xEnVrE:
I also want to stress the fact that with the actual code of the `controlboard`, while the decoupling between coupled and controlled joints will be fine, the joints limits used to instantiate the trajectory generator in the coupled-joints space will be wrong as longs as #472 this is not handled. As a result, the overall range of motion of the fingers will be different from the expected.

ale-git:
Link to geometry documentation:

https://github.com/icub-tech-iit/dev/blob/master/projects/icub-hand-Mk3/calculations/FingerkinV2.docx

Link to the URDF model:

https://github.com/icub-tech-iit/dev/tree/master/projects/icub-hand-Mk3/simulation_model/gazebo/iCubHandMK3_model

traversaro:
> 
> 
> I also want to stress the fact that with the actual code of the `controlboard`, while the decoupling between coupled and controlled joints will be fine, the joints limits used to instantiate the trajectory generator in the coupled-joints space will be wrong as longs as #472 this is not handled. As a result, the overall range of motion of the fingers will be different from the expected.

I guess a possible workaround for now is to pass really loose limits, to ensure that the trajectory generation is not affected?

ale-git:
Although scary, the relation is pretty almost linear:

prox2dist

traversaro:
> 
> 
> Link to geometry documentation:
> 
> https://github.com/icub-tech-iit/dev/blob/master/projects/icub-hand-Mk3/calculations/FingerkinV2.docx
> 
> Link to the URDF model:
> 
> https://github.com/icub-tech-iit/dev/tree/master/projects/icub-hand-Mk3/simulation_model/gazebo/iCubHandMK3_model

I am  not sure if everyone has access to this private repos. In any case, adding these models to public repos is tracked in https://github.com/robotology/icub-model-generator/issues/130, so eventually we could  add the links suggested by @randaz81 .

traversaro:
Just to avoid confusion, the only comment left is https://github.com/robotology/gazebo-yarp-plugins/pull/469#discussion_r391073021 .

plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -258,6 +258,12 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
                 m_coupling_handler.push_back(cpl);
                 yInfo() << "using cer_hand_control";
             }
+            else if (coupling_bottle->get(0).asString()=="icub_hand_mk3")
+            {
+                BaseCouplingHandler* cpl = new HandMk3CouplingHandler(m_robot,coupled_joints, coupled_joint_names);
+                m_coupling_handler.push_back(cpl);
+                yInfo() << "using left_hand_control";
```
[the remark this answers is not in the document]
traversaro: Friendly ping @ale-git

traversaro:
Thanks @ale-git ! @xEnVrE it is ok if I now merge this PR and then you rebase https://github.com/robotology/gazebo-yarp-plugins/pull/499 on the top of the merged `devel`  ?

xEnVrE:
>  @xEnVrE it is ok if I now merge this PR and then you rebase #499 on the top of the merged devel ?

Yep!

Issue #470, opened 2020-03-06
Move gazebo_yarp_controlboard coupling logic to dynamically loaded plugins 


traversaro:
Until now, the coupling logic for the iCub and R1 has been  hardcoded in the `gazebo_yarp_controlboard` plugin, in particular as child classes of the `BaseCouplingHandler` in the file https://github.com/robotology/gazebo-yarp-plugins/blob/3fa2c45a4b8a63ef15c524462124b505f29b3ed9/plugins/controlboard/include/yarp/dev/ControlBoardDriverCoupling.h . As the gazebo-yarp-plugins is meant to be robot-indipendent, this is clearly not ideal, as any new coupled mechanism needs to manually modify the source code of gazebo-yarp-plugins to simulate in Gazebo, see for example https://github.com/robotology/gazebo-yarp-plugins/pull/469 . 
I think that the best solution is to add support in `gazebo_yarp_controlboard` to load coupling logic from classes loaded dynamically. In particular, as the rest of the infrastructure already uses a lot of YARP logic, I think the easiest option is to install the header that contains the `BaseCouplingHandler`, and then permit to downstream users to define YARP device drivers that inherit from `BaseCouplingHandler`, and then specify the name of the YARP device driver to load in the `.ini` configuration file of the `gazebo_yarp_controlboard` plugin. 

In particular, I think we should still use the coupling configuration format (see https://github.com/robotology/icub-gazebo/blob/ab22fdc34c0a9faa6f72428c2abdb6afc3145872/icub/conf/gazebo_icub_left_hand_index.ini#L17): 
~~~
[COUPLING]
index_control (0 1 2) (l_index_proximal l_index_distal reserved)
~~~
But the first string of the line, instead of being an hardcoded string checked in https://github.com/robotology/gazebo-yarp-plugins/blob/bd884ac1ea3a51eaaa448ac92b34782cf9b9ea45/plugins/controlboard/src/ControlBoardDriver.cpp#L219, should be the name of a yarp device that must inherit from the `BaseCouplingHandler` interface. 

cc @fiorisi @randaz81 @ale-git @xEnVrE

PR #472, opened 2020-03-26
Fix Control Board wrong joint limits for coupled joints

### LINKED ISSUE #348: Control Board wrong joint limits for fingers
The `.ini` configuration files for fingers, e.g. [gazebo_icub_left_hand_index.ini](https://github.com/robotology/icub-gazebo/blob/master/icub/conf/gazebo_icub_left_hand_index.ini) contain a group `[LIMITS]` that specifies limits for each __physical__ joint. These limits are loaded within the control board plugin and used to, e.g., specify constraints on the generation of trajectories (see [here](https://github.com/robotology/gazebo-yarp-plugins/blob/5f595c10321b7afae4cc4c89f23fc6c8c6a23493/plugins/controlboard/src/ControlBoardDriver.cpp#L394)).

Since trajectories are generated for each controlled DoF of the finger and not for each physical DoF, e.g. for the index `{l,r}_index_proximal` and ` {l,r}_index_distal`, but the variable `m_jointPosLimits` is not updated to take into account coupling then joint limits used by the trajectory generator are wrong.

Could it be an option to add a method like 

````
decouplePosLimits(std::vector<Range> &posLimits)
````
to the decoupling handlers `ThumbCouplingHandler`, `IndexCouplingHandler`, `MiddleCouplingHandler`, `PinkyCouplingHandler`

and call it in the method [`GazeboYarpControlBoardDriver::setMinMaxPos`](https://github.com/robotology/gazebo-yarp-plugins/blob/5f595c10321b7afae4cc4c89f23fc6c8c6a23493/plugins/controlboard/src/ControlBoardDriver.cpp#L629)?

Or is it better to change the meaning of the section `[LIMITS]` so that it contains the limits of the controlled DoFs instead that of all the physical DoFs?

### PR DESCRIPTION
**Pull request by xEnVrE**
This PR fixes #348 

I had to move the private struct `yarp::dev::GazeboYarpControlBoardDriver::Range` in a separate header file as it is required both within class `yarp::dev::GazeboYarpControlBoardDriver` and  in the signature of a new method in classes `BaseCouplingHandler` and derived.

cc @randaz81 @kouroshD @traversaro

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ### Added
 - Add the possibility to create different types of joints with the `linkattacher` plugin (https://github.com/robotology/gazebo-yarp-plugins/pull/461).
+- Add the possibility to decouple joint limits using method BaseCouplingHandler::decouplePosLimits(), within the 'controlboard' plugin (https://github.com/robotology/gazebo-yarp-plugins/pull/472).
+
+### Fixed
+- Fix evaluation of joints position limits for physically coupled DoFs in method GazeboYarpControlBoardDriver::setMinMaxPos(), within the 'controlboard' plugin (https://github.com/robotology/gazebo-yarp-plugins/pull/472).
 
 ## [3.3.0] - 2019-12-13
 
--- a/plugins/controlboard/CMakeLists.txt
+++ b/plugins/controlboard/CMakeLists.txt
@@ -34,7 +34,8 @@ set(controlBoard_source     src/ControlBoard.cc
 set(controlBoard_headers    include/gazebo/ControlBoard.hh
                             include/yarp/dev/ControlBoardDriver.h
                             include/yarp/dev/ControlBoardDriverTrajectory.h
-                            include/yarp/dev/ControlBoardDriverCoupling.h)
+                            include/yarp/dev/ControlBoardDriverCoupling.h
+                            include/yarp/dev/ControlBoardDriverRange.h)
                             
                             
                             
--- a/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
+++ b/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
@@ -27,6 +27,7 @@
 #include <boost/shared_ptr.hpp>
 #include <ControlBoardDriverTrajectory.h>
 #include <ControlBoardDriverCoupling.h>
+#include <ControlBoardDriverRange.h>
 
 #include <gazebo/common/PID.hh>
 #include <gazebo/common/Time.hh>
@@ -347,12 +348,6 @@ private:
         JointType_Unknown = 0,
         JointType_Revolute,
         JointType_Prismatic
-    };
-
-    struct Range {
-        Range() : min(0), max(0){}
-        double min;
-        double max;
     };
 
     std::string m_deviceName;
--- a/plugins/controlboard/include/yarp/dev/ControlBoardDriverCoupling.h
+++ b/plugins/controlboard/include/yarp/dev/ControlBoardDriverCoupling.h
@@ -9,6 +9,8 @@
 
 #include <gazebo/physics/Model.hh>
 
+#include <ControlBoardDriverRange.h>
+
 namespace yarp {
     namespace dev {
         enum CouplingType
@@ -36,6 +38,8 @@ public:
     virtual bool decoupleAcc(yarp::sig::Vector& current_acc) = 0;
     virtual bool decoupleTrq(yarp::sig::Vector& current_trq) = 0;
 
+    virtual bool decouplePosLimits(std::vector<Range>& pos_limits) = 0;
+
     virtual yarp::sig::VectorOf<int> getCoupledJoints();
     virtual std::string getCoupledJointName (int joint);
     virtual bool checkJointIsCoupled(int joint);
@@ -57,6 +61,8 @@ public:
     bool decoupleAcc (yarp::sig::Vector& current_acc);
     bool decoupleTrq (yarp::sig::Vector& current_trq);
 
+    bool decouplePosLimits(std::vector<Range>& pos_limits);
+
     yarp::sig::Vector decoupleRefPos (yarp::sig::Vector& pos_ref);
     yarp::sig::Vector decoupleRefVel (yarp::sig::Vector& vel_ref);
     yarp::sig::Vector decoupleRefTrq (yarp::sig::Vector& trq_ref);
@@ -73,6 +79,8 @@ public:
     bool decoupleVel (yarp::sig::Vector& current_vel);
     bool decoupleAcc (yarp::sig::Vector& current_acc);
     bool decoupleTrq (yarp::sig::Vector& current_trq);
+
+    bool decouplePosLimits(std::vector<Range>& pos_limits);
 
     yarp::sig::Vector decoupleRefPos (yarp::sig::Vector& pos_ref);
     yarp::sig::Vector decoupleRefVel (yarp::sig::Vector& vel_ref);
@@ -91,6 +99,8 @@ public:
     bool decoupleAcc (yarp::sig::Vector& current_acc);
     bool decoupleTrq (yarp::sig::Vector& current_trq);
 
+    bool decouplePosLimits(std::vector<Range>& pos_limits);
+
     yarp::sig::Vector decoupleRefPos (yarp::sig::Vector& pos_ref);
     yarp::sig::Vector decoupleRefVel (yarp::sig::Vector& vel_ref);
     yarp::sig::Vector decoupleRefTrq (yarp::sig::Vector& trq_ref);
@@ -107,6 +117,8 @@ public:
     bool decoupleVel (yarp::sig::Vector& current_vel);
     bool decoupleAcc (yarp::sig::Vector& current_acc);
     bool decoupleTrq (yarp::sig::Vector& current_trq);
+
+    bool decouplePosLimits(std::vector<Range>& pos_limits);
 
     yarp::sig::Vector decoupleRefPos (yarp::sig::Vector& pos_ref);
     yarp::sig::Vector decoupleRefVel (yarp::sig::Vector& vel_ref);
@@ -125,6 +137,8 @@ public:
     bool decoupleAcc (yarp::sig::Vector& current_acc);
     bool decoupleTrq (yarp::sig::Vector& current_trq);
 
+    bool decouplePosLimits(std::vector<Range>& pos_limits);
+
     yarp::sig::Vector decoupleRefPos (yarp::sig::Vector& pos_ref);
     yarp::sig::Vector decoupleRefVel (yarp::sig::Vector& vel_ref);
     yarp::sig::Vector decoupleRefTrq (yarp::sig::Vector& trq_ref);
@@ -142,6 +156,8 @@ public:
     bool decoupleAcc (yarp::sig::Vector& current_acc);
     bool decoupleTrq (yarp::sig::Vector& current_trq);
 
+    bool decouplePosLimits(std::vector<Range>& pos_limits);
+
     yarp::sig::Vector decoupleRefPos (yarp::sig::Vector& pos_ref);
     yarp::sig::Vector decoupleRefVel (yarp::sig::Vector& vel_ref);
     yarp::sig::Vector decoupleRefTrq (yarp::sig::Vector& trq_ref);
@@ -158,6 +174,8 @@ public:
     bool decoupleVel (yarp::sig::Vector& current_vel);
     bool decoupleAcc (yarp::sig::Vector& current_acc);
     bool decoupleTrq (yarp::sig::Vector& current_trq);
+
+    bool decouplePosLimits(std::vector<Range>& pos_limits);
 
     yarp::sig::Vector decoupleRefPos (yarp::sig::Vector& pos_ref);
     yarp::sig::Vector decoupleRefVel (yarp::sig::Vector& vel_ref);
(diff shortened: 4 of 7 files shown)

### CODE REVIEW

**Comment by traversaro** [comment]
As this is a breaking change, can you please target  the branch `devel` and also update the https://github.com/robotology/gazebo-yarp-plugins/blob/devel/CHANGELOG.md  ? Thanks!

**Comment by xEnVrE** [comment]
> As this is a breaking change, can you please target the branch `devel` and also update the https://github.com/robotology/gazebo-yarp-plugins/blob/devel/CHANGELOG.md ? Thanks!

Sure!

File: CHANGELOG.md [hunk 1]
```diff
@@ -7,6 +7,10 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ### Added
 - Add the possibility to create different types of joints with the `linkattacher` plugin (https://github.com/robotology/gazebo-yarp-plugins/pull/461).
+- Add the possibility to decouple joint limits using method BaseCouplingHandler::decouplePosLimits(), within the 'controlboard' plugin (https://github.com/robotology/gazebo-yarp-plugins/pull/472).
+
+## Fixed
```
traversaro:
```suggestion
### Fixed
```

xEnVrE:
> One thing I do not get, but probably I am missing something: if https://github.com/robotology/gazebo-yarp-plugins/pull/472/files#diff-a01901eeedeed61dae3fabb62c01e331R689 is just applying the existing position decoupling rules to the min and max vectors, can't we use the existing decouplePos function on two temporary min and max vectors to get the same result, without duplicating the decoupling position code in multiple methods?

I think it is doable. It was implemented with additional methods because that what was agreed in https://github.com/robotology/gazebo-yarp-plugins/issues/348 and also to exploit the existing  `std::vector<Range>`. 

Of course, we can change as you suggested by extracting two temporary vectors from `m_jointPosLimits`, applying `decouplePos` and then copying values back.

CHANGELOG.md
xEnVrE: Done!

traversaro:
> Of course, we can change as you suggested by extracting two temporary vectors from m_jointPosLimits, applying decouplePos and then copying values back.

I was thinking  that probably it make sense to do so, instead of having a  lot of decouplePosLimits that are not implemented.

xEnVrE:
> I was thinking that probably it make sense to do so, instead of having a lot of decouplePosLimits that are not implemented, to be honest I had completely forgotten the discussion in #348 .

Fine with me. I will change it then.

xEnVrE:
> > I was thinking that probably it make sense to do so, instead of having a lot of decouplePosLimits that are not implemented, to be honest I had completely forgotten the discussion in #348 .
> 
> Fine with me. I will change it then.

After some checks on the simulator, I am not sure that this solution will work for all the couplings. In particular, I think that there is something wrong with the handling of both the `eyes` and `fingers abduction`. However, it is not easy to understand what is going on since:
- the implementation of the coupling handlers of `eyes` and `fingers abduction` is not clear to me;
- the limits of controlled DoFs (i.e. controlled inside `Gazebo`) seems wrong to me.

Hence, I propose to freeze this PR for now. Meanwhile I will open issues here and on `icub-gazebo` to clarify these aspects.

traversaro:
> Hence, I propose to freeze this PR for now. Meanwhile I opened issues #473 and robotology/icub-gazebo#75 to clarify these aspects.

Great, thanks.

traversaro:
I wonder if also this discussion is related: https://github.com/robotology/gazebo-yarp-plugins/pull/324 .

xEnVrE:
> I wonder if also this discussion is related: #324 .

In this issue, at some point @triccyx says (taken from https://github.com/robotology/gazebo-yarp-plugins/pull/324#issuecomment-353297638):

> At the end the file could be correct because it describes the limit of joint 4 and 5 that are Vs and Vg after the coupling line:
[COUPLING]
eyes_vergence_control (4 5) (eyes_version eyes_vergence)

where the *file* is the configuration file associated to the head, i.e. https://github.com/robotology/icub-gazebo/blob/master/icub/conf/gazebo_icub_head.ini

If we try to setup an automatic procedure to infer the limits for coupled joints using the same logic that we use to move from the **physical joint space** to the **controlled DoFs**, i.e. using `decouplePos` as you was suggesting, we need to be sure that limits in the configuration files are always expressed in terms of physical joints. 

As an example, for the index distal joint - check the [configuration file](https://github.com/robotology/icub-gazebo/blob/master/icub/conf/gazebo_icub_right_hand_index.ini) - we have limits for the index proximal and for the two **physical** joints associated with the unique DoF that on the iCub robot is typically identified as `{l, r}_index_distal`. For the eyes, instead, limits are expressed in terms of tilt, version and vergence while, correct me If I am wrong, they should be expressed in terms of tilt, left_pan and right_pan in order to be coherent. An additional issue is that, given the special nature of the relationships linking version and vergence to left/right pan, I do not think that applying `decuplePos` on the left/right pan limits would give us the correct limits for version and vergence.

xEnVrE:
> One thing I do not get, but probably I am missing something: if https://github.com/robotology/gazebo-yarp-plugins/pull/472/files#diff-a01901eeedeed61dae3fabb62c01e331R689 is just applying the existing position decoupling rules to the min and max vectors, can't we use the existing decouplePos function on two temporary min and max vectors to get the same result, without duplicating the decoupling position code in multiple methods?

After some tests with `eyes version and vergence` is seems to me that is not possible to find joint limits by simply applying the decoupling rules to the min and max vectors. I think that with even more complicated coupling rules, e.g. the one implemented in #469, the outcome is the same. I think we need separated methods to handle limits decoupling.

traversaro:
> 
> 
> > One thing I do not get, but probably I am missing something: if https://github.com/robotology/gazebo-yarp-plugins/pull/472/files#diff-a01901eeedeed61dae3fabb62c01e331R689 is just applying the existing position decoupling rules to the min and max vectors, can't we use the existing decouplePos function on two temporary min and max vectors to get the same result, without duplicating the decoupling position code in multiple methods?
> 
> After some tests with eyes version and vergence is seems to me that is not possible to find joint limits by simply applying the decoupling rules to the min and max vectors. I think that with even more complicated coupling rules, e.g. the one implemented in #469, the outcome is the same, i.e. we need specific methods to handle limits decoupling.
> 
> @traversaro

Yes, now  that you wrote it explicitly it make totally sense, even for linear coupling it is not true that if (given $x \in \mathbb{R}^n$):

$$
x_{min} < x < x_{max}
$$ 

then for every matrix $A \in \mathbb{R}^n$ you also have: 
$$
A x_{min} < A x < A x_{max}
$$ 

in particular, this is true only if  $A$ is composed just by positive coefficients. If $A$ is a general matrix, solving the inequality is more tricky (see https://www.jstor.org/stable/1967869 and https://math.stackexchange.com/questions/1310079/using-linear-algebra-e-g-matrix-methods-to-solve-a-system-of-linear-inequalit). I wonder how we are handling this at the firmware level on the iCub. 

Given this, I guess we should take a step back, and try to understand in which way are are actually using joint limits. By doing a rough research ( https://github.com/robotology/gazebo-yarp-plugins/search?q=m_jointPosLimits&unscoped_q=m_jointPosLimits ) it seems that the only place in which the limits are used is in the clipping of the generated trajectories: 
https://github.com/robotology/gazebo-yarp-plugins/blob/06629530dfd5e4ec66287515ea61a9af750de6b7/plugins/controlboard/src/ControlBoardDriverTrajectory.cpp#L223 and then the limits are exposed to the downstream user: 
https://github.com/robotology/gazebo-yarp-plugins/blob/01ad1d1252b0827033fd57058f066cc790b893c6/plugins/controlboard/src/ControlBoardDriverOthers.cpp#L46

So, to be honest I do not remember if the trajectory generation is happening at the level of coupled quanties, or using the motor/physical joints (@xEnVrE if you can provide this info it would be great), so there are two cases: 
* Joint trajectory generation is done with the coupled quantities: In that case everything works fine, and we should just clarify that the joint limits specified by the user in the `.ini` file refer to the **coupled quantities** (so version and vergence for the eyes, not left_pan and right_pan) 
* Joint trajectory generation is done with the coupled quantities: in this case, we need both coupled  limits and physical limits. This can be either an appropriate additional method in the coupling classes such as the one proposed in this PR, or the user could specify both the coupled and the physical limits, and we will trust him to provide coherent data.

traversaro:
> Joint trajectory generation is done with the coupled quantities: in this case, we need both coupled limits and physical limits. This can be either an appropriate additional method in the coupling classes such as the one proposed in this PR, or the user could specify both the coupled and the physical limits, and we will trust him to provide coherent data.

Actually that is much problematic then expected, as the solution for a system of inequalities:
$$
Ax > b
$$
do not always have a solution in the form: 
$$
x_{min} < x < x_{max} 
$$

so it is simply not possible to pass from one to another, unless you explicitly do not consider  possible valid configuration.

For a quick example, think of: 

$$
A x > 0
$$
with: 
$$
A = 
\begin{bmatrix} 
1 & 1 \\
1 & -1 
\end{bmatrix}
$$
and 
$$
x 
=
\begin{bmatrix} 
x_1 \\
x_2
\end{bmatrix}
$$

The solution is: 
$$
x_1 > 0
x_1 > x_2 
$$

xEnVrE:
Hi @traversaro,

thank you for the analysis! You are totally right. In fact, for my experiments with eyes, I had to manually find max and mins for eyes left and right pan given limits on eyes vergence and version that we can take from robot hardware configurations. I then inserted the evaluated limits for pans on the configuration file of the plugin, while I had to put limits for eyes version and vergence inside the  newly introduced `EyesCouplingHandler::decouplePosLimits` (of course, not the best solution as the plugin is, in an ideal world, decoupled from the configuration file of the robot is used with).

To recap:
- from my understanding, control is done using actual implemented joints on the model, e.g. for eyes left_pan and right_pan, while trajectories are generated for the coupled-joints space, e.g. version and vergence.
- limits inside the configuration files have to be in the implemented-joints space, because those configuration files always contained information (like PID, etc) for each actually implemented joint in the Gazebo model. That is way I was surprised to find that, see https://github.com/robotology/icub-gazebo/issues/75 and https://github.com/robotology/icub-gazebo/blob/14a078e49968c2130d64b8d5635d6f302035d67b/icub/conf/gazebo_icub_head.ini#L72-L75, limits for eyes were specified at version/vergence level instead of pan left / pan right
- limits must be converted to the coupled-joints space inside the `controlboard` plugin

Relative to the last points, trajectories limits are configured in two places:

https://github.com/robotology/gazebo-yarp-plugins/blob/81f7dc105d95a989c7a677a2848e32c398a6f7e4/plugins/controlboard/src/ControlBoardDriver.cpp#L383-L386

and

https://github.com/robotology/gazebo-yarp-plugins/blob/81f7dc105d95a989c7a677a2848e32c398a6f7e4/plugins/controlboard/src/ControlBoardDriver.cpp#L383-L386

That is way, in this PR, I proposed to modify method `setMinMaxPos` that is called here

https://github.com/robotology/gazebo-yarp-plugins/blob/81f7dc105d95a989c7a677a2848e32c398a6f7e4/plugins/controlboard/src/ControlBoardDriver.cpp#L290

in order to modify quantities `m_jointPosLimits` so that they contain the right limits for coupled joints.

xEnVrE:
> or the user could specify both the coupled and the physical limits, and we will trust him to provide coherent data

I like more this idea than the solution implemented in this PR. This way, the behavior will depend only on the user data that is confined in the configuration file. Instead, at the moment, we might be obliged to have some of the limits in the configuration files and part of the logic in this plugin. Considering that the plugin and the configuration files are in different repositories, I don't really like this solution.

traversaro:
Ok, thanks @xEnVrE , now everything is crystal  clear to me. So the point is that the answer to my question in https://github.com/robotology/gazebo-yarp-plugins/pull/472#pullrequestreview-383016448 "can't we use the existing decouplePos function on two temporary min and max vectors to get the same result?" is simply "no", or more in detail: "no, because that works fine only for couplings for which all the coefficients are positive", right?

xEnVrE:
Right! 

Nevertheless, I am more for the solution where the user specifies both limits in the configuration file shipped within a specific repository containing models of a robot.

The best solution would be to follow the path you proposed in https://github.com/robotology/gazebo-yarp-plugins/issues/470. This way, correct me If  I am wrong, we can move code for coupling handlers inside the repository of robots models such that:
- coupling plugin loading mechanism is in `gazebo-yarp-plugins`
- configuration file and code for coupling specific to the robot are inside the repository of the robot and not in `gazebo-yarp-plugins`

traversaro:
> This way, correct me If I am wrong, we can move code for coupling handlers inside the repository of robots models such that:

Exactly, that was the idea. However I think that the modification to correctly handle the limits (either a combination of this PR + modification to load explicitly the coupled axis limits or simply just loading explicitly the coupled axis limits) can be done quite independently from that. If you are  interested in working on that proposal and you need help feel free to let me know, thanks!

traversaro:
@xEnVrE how do you prefer to proceed? We merge this PR for the time being, or you prefer to add instead a new parameter (probably under the `[COUPLING]` group) to specify the coupled joint limits directly?

xEnVrE:
Sorry @traversaro, I had no time to handle this yesterday. I will address it next week and I will add new parameters for coupled joint limits.

traversaro:
> 
> 
> Sorry @traversaro, I had no time to handle this yesterday. I will address it next week and I will add new parameters for coupled joint limits.

No hurry, I just wanted to be sure I was not blocking anyone here.

traversaro:
Hi @xEnVrE, just to understand, are you still interested in work on this in the short term?

xEnVrE:
Hi @traversaro, definitively. If you want, just to clean things a bit, we can close this PR and I will open a new one since the solution we agreed on is different from the original one.

traversaro:
> 
> 
> Hi @traversaro, definitively. If you want, just to clean things a bit, we can close this PR and I will open a new one since the solution we agreed on is different from the original one.

Yes, probably this is a good idea!

PR #481, opened 2020-04-27
[videotexture] fixed image conversion

pattacini:
This PR attempts to fix a compilation error happening while building the `videotexture` plugin.
Image conversion is now done through the standard `yarp::cv::toCvMat` function.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ set(${PROJECT_NAME}_VERSION
 include(GNUInstallDirs)
 
 # Finding dependencies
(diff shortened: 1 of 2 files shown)


traversaro:
Just to understand, on which system the compilation is failing?

pattacini:
Ops, didn't see your comment and just merged.
Was failing on Ubuntu 20.04

traversaro:
> Ops, didn't see your comment and just merged.
> Was failing on Ubuntu 20.04

Yes,  actually I noticed that this PR is making the dependency on OpenCV compulsory, but I will fix this in a subsequent PR.

PR #484, opened 2020-05-04
Make OpenCV optional again

traversaro:
OpenCV has always been optional for gazebo-yarp-plugins, but it was made a required dependency (transitively by depending on YARP's cv component) in  https://github.com/robotology/gazebo-yarp-plugins/pull/481 .

This PR makes the OpenCV dependency optional again.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,17 @@ set(${PROJECT_NAME}_VERSION
 include(GNUInstallDirs)
 
 # Finding dependencies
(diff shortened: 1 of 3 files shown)


traversaro:
> 
> 
> The `videotexture` plugin needs to have `YARP::cv` available to build; thus, I'm afraid that it should be put under conditional compilation.

It already is, see https://github.com/robotology/gazebo-yarp-plugins/pull/484/files#diff-5a8faf6e4477daafbcbe7625b3987686R26 .

pattacini:
Since it wasn't a proposed change, I didn't notice it 👍

PR #485, opened 2020-05-05
Avoid useless linking in gazebo_yarp_singleton

traversaro:
`${YARP_LIBRARIES}` has the peculiar behavior that the contained targets depends on the arguments passed `find_package(YARP ...)` in this project. In particular, if (and only if) `GAZEBO_YARP_PLUGINS_HAS_OPENCV` is enabled, then  also the `YARP::YARP_cv` component of yarp is included, that is not required at all (see https://github.com/robotology/gazebo-yarp-plugins/pull/481).

For this reason, it is better to just link the libraries we actually depend on, otherwise it will be impossible to use `gazebo_yarp_singleton` in a downstream project that do not have the  `YARP::YARP_cv` target defined.

An alternative was to call `find_dependency(YARP ... COMPENENTS cv)` in the config file of `GazeboYARPPlugins` , but this does not make a lot of sense as `YARP::YARP_cv` is not actually a public dependency of any exported library.


randaz81:
This PR fixes the issues on CER gazeboTripod plugin

PR #492, opened 2020-05-23
[plugins][controlboard] Fix wrong feedback measurements for coupled joints

### LINKED ISSUE #491: Velocity and Mixed control of coupled joints uses the wrong feedback measurements
Sorry for the long issue but a bit of context is required: the `controlboard` plugin uses a set of methods called `decouple*` and `decoupleRef*` to handle coupled joints:
- `decouple*` goes from actual physical joints to coupled joints (e.g. from eyes left/right pan to eyes version/vergence)
- `decoupleRef*` does the opposite

Typically, `decouple*` is used when we need to initialize the trajectory generators, while `decoupleRef*` is used to generate measurements for closed loop control at Gazebo level.

An example:

https://github.com/robotology/gazebo-yarp-plugins/blob/6275d40549f5226e07847f0f198020416c37d3c6/plugins/controlboard/src/ControlBoardDriver.cpp#L481-L492

`m_motPositions` store the values used to close the loop, i.e. those of `m_positions`, and, after that, `m_positions` is overwritten with values generated by `decouplePos` that are used in 

https://github.com/robotology/gazebo-yarp-plugins/blob/6275d40549f5226e07847f0f198020416c37d3c6/plugins/controlboard/src/ControlBoardDriverPositionControl.cpp#L19

to initialize the trajectory generator when `positionMove` is called from the user side.

Finally, `m_motPositions` is used to close the loop in

https://github.com/robotology/gazebo-yarp-plugins/blob/6275d40549f5226e07847f0f198020416c37d3c6/plugins/controlboard/src/ControlBoardDriver.cpp#L573

## Issue with velocity control

In few words, the loop at Gazebo level is closed using references from the actual physical joints and measurements from the coupled-joints world. E.g., references of eyes left and right pan and measurements of eyes version and vergence. Of course, this is wrong.

### Detailed explanation

In case of velocity control, there is not a variable similar to `m_motPositions`, i.e., `m_motVelocities`, but only a variable called `m_velocities`. 

https://github.com/robotology/gazebo-yarp-plugins/blob/6275d40549f5226e07847f0f198020416c37d3c6/plugins/controlboard/src/ControlBoardDriver.cpp#L481-L492

`m_velocities` is replaced, in case of coupled joints, with measurements from the coupled-joints world, e.g. eyes version and vergence velocities, and, later,

https://github.com/robotology/gazebo-yarp-plugins/blob/6275d40549f5226e07847f0f198020416c37d3c6/plugins/controlboard/src/ControlBoardDriver.cpp#L583

is used to close the loop against `m_motReferenceVelocities` that, instead, is the reference for actual physical joints, e.g. eyes pan left and right. Indeed, `m_motReferenceVelocities`, in case of coupled joints, is set here

https://github.com/robotology/gazebo-yarp-plugins/blob/6275d40549f5226e07847f0f198020416c37d3c6/plugins/controlboard/src/ControlBoardDriver.cpp#L560

### Possible solutions:

- at the moment, it seems that the values generated by `decoupleVel(m_velocities)` are not used to generate trajectories because the `RampFilter` for velocity generation only needs the reference velocity 

https://github.com/robotology/gazebo-yarp-plugins/blob/6275d40549f5226e07847f0f198020416c37d3c6/plugins/controlboard/src/ControlBoardDriverVelocityControl.cpp#L22

Then, we could simply remove the line 
```
m_coupling_handler[cpl_cnt]->decoupleVel(m_velocities);
```

- another, more clean solution, is to add a `m_motVelocities` variable and use that to close the loop

##  Issue with mixed mode control

As done for position control

https://github.com/robotology/gazebo-yarp-plugins/blob/6275d40549f5226e07847f0f198020416c37d3c6/plugins/controlboard/src/ControlBoardDriver.cpp#L573

we should use `m_motPositions` to close the loop. 

Instead, in case of `VOCAB_CM_MIXED`, 

https://github.com/robotology/gazebo-yarp-plugins/blob/6275d40549f5226e07847f0f198020416c37d3c6/plugins/controlboard/src/ControlBoardDriver.cpp#L594

`m_positions` is used.

I can open a PR to fix the issues.

cc @traversaro

xEnVrE:
Within this PR, several fixes for control modes in presence of coupled joints:

- add a variable `m_motVelocities` alongside `m_velocities`: `m_velocities` contains the feedback in the coupled-joints space, while `m_motVelocities` have to be used instead to close the loop as it contains the measurements in the implemented-joint space
- use `m_motVelocities` instead of `m_velocities` to close the loop
- use `m_motPositions` instead of `m_positions` to close the loop

**Note:** for position and mixed control modes, a variable `m_motPositions` was already present but, in some places, `m_positions` was still used to close the loop.

Fixes #491

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ## [Unreleased]
 
+### Fixed
+- Fixed wrong measurements feedback used for coupled joints in `gazebo_yarp_controlboard`(https://github.com/robotology/gazebo-yarp-plugins/pull/492).
+
 ## [3.4.0] - 2020-05-19
 
 ### Added
--- a/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
+++ b/plugins/controlboard/include/yarp/dev/ControlBoardDriver.h
@@ -385,6 +385,7 @@ private:
 
     yarp::sig::Vector m_positions;          /**< joint positions [Degrees] */
     yarp::sig::Vector m_motPositions;      /**< motor positions [Degrees] */
+    yarp::sig::Vector m_motVelocities;      /**< motor velocities [Degrees/Seconds] */
     yarp::sig::Vector m_velocities;         /**< joint velocities [Degrees/Seconds] */
     yarp::sig::Vector m_torques;            /**< joint torques [Netwon Meters] */
     yarp::sig::Vector m_measTorques;        /**< joint torques from virtual analog sensor [Newton Meters] */
--- a/plugins/controlboard/src/ControlBoardDriver.cpp
+++ b/plugins/controlboard/src/ControlBoardDriver.cpp
@@ -62,6 +62,7 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
     m_zeroPosition.resize(m_numberOfJoints);
     m_jntReferenceVelocities.resize(m_numberOfJoints);
     m_velocities.resize(m_numberOfJoints);
+    m_motVelocities.resize(m_numberOfJoints);
     m_amp.resize(m_numberOfJoints);
     m_torques.resize(m_numberOfJoints); m_torques.zero();
     m_measTorques.resize(m_numberOfJoints); m_measTorques.zero();
@@ -101,6 +102,7 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
     m_motPositions.zero();
     m_zeroPosition.zero();
     m_velocities.zero();
+    m_motVelocities.zero();
     m_motReferencePositions.zero();
     m_motReferenceVelocities.zero();
     m_motReferenceTorques.zero();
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ## [Unreleased]
 
+## [3.4.1] - 2020-05-23
```
traversaro, line 9:
```suggestion
```

Leaving this comments under `## [Unreleased]` is ok. We will add a version number (and a release date) when an actual release will be done.

traversaro:
Thanks @xEnVrE, this seems great to me! 
I  would wait for some days to see if @randaz81 or @kouroshD has any comment on this, if they do not have any objection I think we can merge.

CHANGELOG.md
xEnVrE: Done!

traversaro:
Thanks @xEnVrE .

PR #502, opened 2020-07-21
[plugins][maissensor] Fix discrepancy of MAIS analog sensor between simulation and real robot

### LINKED ISSUE #476: Fix the discrepancy of the output IAnalogSensor device between real robot and gazebo-yarp-plugins
Currently, the outputs of the IAnalogSensor Device from the real robot are the raw values, i.e., 0-255 values.  However, the output of simulated robot are in radians. This behavior has been observed for the values of the analog sensors of the hand joints.

CC @traversaro

xEnVrE:
Fixed mismatched behavior, with respect to the real robot, of MAIS analog sensors output

Output is constrained to be between 0 and 255, the former is reached when the joint is >= maximum value, the latter when the joint is <= its minimum value

Maximum and minimum values are obtained from Gazebo hw limits.

Fixes #476

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 ### Fixed
 - Fixed wrong measurements feedback used for coupled joints in `gazebo_yarp_controlboard`(https://github.com/robotology/gazebo-yarp-plugins/pull/492).
 - Fixed mismatched behavior, with respect to the real robot, of coupling handler FingersAbductionCouplingHandler in 'gazebo_yarp_controlboard' plugin (https://github.com/robotology/gazebo-yarp-plugins/pull/499)
+- Fixed mismatched behavior, with respect to the real robot, of mais analog sensors in 'gazebo_yarp_maissensor' plugin (https://github.com/robotology/gazebo-yarp-plugins/pull/500)
 
 ### Added
 - Add external wrench smoothing feature for `externalwrench` plugin (https://github.com/robotology/gazebo-yarp-plugins/pull/495)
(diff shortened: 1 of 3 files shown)


pattacini:
Can we revise these modifications in such a way that they can be made configurable?
We need to fix the MAIS handling on iCub quite urgently.

xEnVrE:
> Can we revise these modifications in such a way that they can be made configurable?

If ok, I can add a parameter to specify the format of the output. And, not to break the current status, to have by default the output in degrees.

I though that only icub robots were using this plugin also because there was a hardcoded

https://github.com/robotology/gazebo-yarp-plugins/blob/690f22fed7b994561e47a9b2331465f04b21e924/plugins/maissensor/src/MaisSensorDriver.cpp#L58

which, btw, I changed to 15 within this PR as that is the number of channels we are expecting on icub robots (according to the [icub wiki](http://wiki.icub.org/wiki/Software_interface:_standard_port_names_for_hardware_devices#Finger_encoders)).

randaz81:
If urgent, a parameter is ok as a temporary fix because, unfortunately, I think we need some time to investigate an optimal solution.
Other possibilities could be, just for example, a conversion performed in the wrapper layer (this kind of operation is typically done in the RawToUser layer of yarp), or a modification of the CER model (using a different/new plugin and not the maisSensorDriver).
PS: I also have a bad feeling that using 0-255 values in icub, will break some ROS compatibility, but I'm not sure if this feature is used by someone.

traversaro:
Thanks @xEnVrE for the PR and @randaz81 for the review. @xEnVrE if you can resolve the CHANGELOG conflict, then I can merge, thanks!

randaz81:
After some additional reasoning, I come out with the idea that the fixes suggested by Nicola are right: the plugin in gazebo must reflect as much as possible the implementation on the real hardware. In this case iCub is ok, while R1 implementation is contrived and it needs some refactoring. Please proceed with this PR, I'll take care of the fixes required on R1.

Issue #240, opened 2016-05-05
Velocity joint control does not work properly


plinioMoreno:
@randaz81 and @traversaro I have been able to control joints in position and mixed mode, but in velocity mode the following behavior happens:

If the motion of the joint is affected by gravity (the motor axis is parallel to the world plane), the joint is not able to counteract gravity and moves towards the limit.
If the motion of the joint is not affected by gravity (the motor axis is parallel to the gravity vector), the joint oscillates between the minimum and maximum limits.

The parameters I'm using are as follows:

```
[VELOCITY_CONTROL]
controlUnits  metric_units
controlLaw    joint_pid_gazebo_v1
kp           20.0  20.0 20.0  20.0  20.0
kd           0.0   0.0  0.0   0.0   0.0
ki           0.01  0.01 0.01  0.01  0.01
maxInt       9999  9999 9999  9999  9999
maxOutput    9999  9999 9999  9999  9999
shift        0     0    0     0     0
ko           0     0    0     0     0
stictionUp   0     0    0     0     0
stictionDwn  0     0    0     0     0
```

And from the yarpmotorgui I get the following output:

```
[INFO]joint: 2 in VELOCITY mode
[INFO]Changing reference acceleration of joint  2  to 100000 
[INFO]joint: 2 in POSITION mode
[INFO]joint: 0 in VELOCITY mode
[INFO]Changing reference acceleration of joint  0  to 100000
```

Let me know if you have any suggestion, and I wonder why the acceleration is set so high

plinioMoreno:
Hi again,

Additional information about the tests I've done:

- I tested several PID values and the behaviour is always the same. Also played with the shift to see if I was scaling badly the parameters, but always get the same behavior.

- I am aiming at controlling Vizzy's [head motors](https://github.com/vislab-tecnico-lisboa/vizzy/tree/master/vizzy_description/urdf/head). All the files head.yarp.* are the ones loaded. I am using Gazebo 7.0.

- I tried also using the `simpleClient` executable, where I can control all the values (especially the acceleration that is so high when I use yarpmotorgui), but I get always the same behavior.

Any suggestion would be very appreciated.

traversaro:
Sorry for the late answer @plinioMoreno! 
As far I remember the velocity control on the Gazebo simulator uses directly the Gazebo JointController, that closes a velocity PID directly on the measured velocity, while the mixed mode and the velocity control on the real robot actually use a position loop underneath the velocity control loop. 

[...]

PR #514, opened 2020-09-21
Add velocityControlImplementationType option to gazebo_yarp_controlboard plugin

traversaro:
This option permits to switch between using a velocity PID for the Velocity Control Mode (what has been implemented until now) or the use of an integrator that integrates the velocity reference and then uses the position low level PID.

The setting is now optional and if not present will default to `direct_velocity_pid`, but it will be compulsory in gazebo-yarp-plugins 4.x .

Fix https://github.com/robotology/gazebo-yarp-plugins/issues/240 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


pattacini (approved):
LGTM 🚀

PR #517, opened 2020-10-02
Fix crash when gazebo_yarp_controlboard is closed

traversaro:
This is a regression introduced in https://github.com/robotology/gazebo-yarp-plugins/pull/403 . When a `gazebo_yarp_controlboard`  was closed, a pointer was checked if it was 0, but if the configuration flag `useVirtualAnalogSensor` was set to false, the pointer was not initialized, so it could be different from 0 even if it was not a valid pointer.

### DIFF
--- a/plugins/controlboard/src/ControlBoard.cc
+++ b/plugins/controlboard/src/ControlBoard.cc
@@ -24,7 +24,7 @@ namespace gazebo
 GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
 
     GazeboYarpControlBoard::GazeboYarpControlBoard() : m_iWrap(0),
-                                                       m_iVirtAnalogSensorWrap(0)
+                                                       m_iVirtAnalogSensorWrap(nullptr)
     {}
 
     GazeboYarpControlBoard::~GazeboYarpControlBoard()
--- a/plugins/controlboard/src/ControlBoard.cc
+++ b/plugins/controlboard/src/ControlBoard.cc
@@ -23,7 +23,8 @@ namespace gazebo
 
 GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
 
-    GazeboYarpControlBoard::GazeboYarpControlBoard() : m_iWrap(0)
+    GazeboYarpControlBoard::GazeboYarpControlBoard() : m_iWrap(0),
+                                                       m_iVirtAnalogSensorWrap(0)
     {}
 
     GazeboYarpControlBoard::~GazeboYarpControlBoard()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ### Fixed 
 - Fixed use of the `VOCAB_CM_MIXED` control mode when the physics timestep is different from 1 millisecond (https://github.com/robotology/gazebo-yarp-plugins/pull/514).
+- Fixed missing initialization of a pointer in `gazebo_yarp_controlboard` . In some cases this was causing crashes when a model that contained a `gazebo_yarp_controlboard` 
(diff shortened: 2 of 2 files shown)


prashanthr05 (approved):

prashanthr05:
> This is a regression introduced in #403 . When a `gazebo_yarp_controlboard` was closed, a pointer was checked if it was 0, but if the configuration flag `useVirtualAnalogSensor` was set to false, the pointer was not initialized, so it could be different from 0 even if it was not a valid pointer.

Ok, this was the reason that Gazebo crashed every time the model was deleted?

traversaro:
> 
> 
> > This is a regression introduced in #403 . When a `gazebo_yarp_controlboard` was closed, a pointer was checked if it was 0, but if the configuration flag `useVirtualAnalogSensor` was set to false, the pointer was not initialized, so it could be different from 0 even if it was not a valid pointer.
> 
> Ok, this was the reason that Gazebo crashed every time the model was deleted?

I have no idea. Do you have an issue tracking the problem of Gazebo crash when a model was deleted?

plugins/controlboard/src/ControlBoard.cc
```diff
@@ -23,7 +23,8 @@ namespace gazebo
 
 GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
 
-    GazeboYarpControlBoard::GazeboYarpControlBoard() : m_iWrap(0)
+    GazeboYarpControlBoard::GazeboYarpControlBoard() : m_iWrap(0),
+                                                       m_iVirtAnalogSensorWrap(0)
```
prashanthr05, line 27: should we change this to `0` to `nullptr` ? here and elsewhere?

traversaro: good point:
```suggestion
                                                       m_iVirtAnalogSensorWrap(nullptr)
```

prashanthr05:
Could you also fix this in this PR? https://github.com/robotology/gazebo-yarp-plugins/pull/403/files#r236188967

prashanthr05:
> Do you have an issue tracking the problem of Gazebo crash when a model was deleted?

Nope! I will open one with the details, the next time I replicate the crash.

traversaro:
> 
> 
> Could you also fix this in this PR? https://github.com/robotology/gazebo-yarp-plugins/pull/403/files#r236188967

I don't see how that is related. If you want to fix that, feel free to propose a PR with that change.

plugins/controlboard/src/ControlBoard.cc
```diff
@@ -24,7 +24,7 @@ namespace gazebo
 GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
 
     GazeboYarpControlBoard::GazeboYarpControlBoard() : m_iWrap(0),
-                                                       m_iVirtAnalogSensorWrap(0)
+                                                       m_iVirtAnalogSensorWrap(nullptr)
```
prashanthr05, line 27: Could you also make relevant changes in the destructor?

prashanthr05:
> > Could you also fix this in this PR? https://github.com/robotology/gazebo-yarp-plugins/pull/403/files#r236188967
> 
> I don't see how that is related. If you want to fix that, feel free to propose a PR with that change.

Understood, I shall open a PR for that.

plugins/controlboard/src/ControlBoard.cc
traversaro: I think that any 0/nullptr cleanup can be done in a separate PR.

prashanthr05: > Could you also make relevant changes in the destructor?


Since this is unrelated to the fix suggested in this PR, I shall open another PR for that.

prashanthr05:
The changes are ok for me, you could merge it once the CI passes. Thank you!

PR #519, opened 2020-10-02
Use nullptr for initialization of pointer objects of wrappers in ControlBoard instead of 0

prashanthr05:
Change constructor initialization and destructor assignment from 0 to `nullptr` for `m_iWrap` and `m_iVirtAnalogSensorWrap`.

Can be rebased merged after #517.


traversaro (approved):

traversaro (approved):

prashanthr05:
Should I update the changelog for this? or is it safe to not update?

traversaro:
> 
> 
> Should I update the changelog for this? or is it safe to not update?

As it is an internal only change that does not affect the users in any way, it is safe not to mention it in the Changelog.

prashanthr05:
> > Should I update the changelog for this? or is it safe to not update?
> 
> As it is an internal only change that does not affect the users in any way, it is safe not to mention it in the Changelog.

Ok, thanks!

PR #527, opened 2021-01-12
Add trapezoidal speed trajectory generator

PeterBowman:
The `TrapezoidalSpeedTrajectoryGenerator` class is added for use by `GazeboYarpControlBoardDriver`. It is meant to generate a trapezoidal profile capped at the provided reference speed and using fixed acceleration in both ramps. The signature of `TrajectoryGenerator::initTrajectory` has been expanded to accomodate said reference acceleration value. This generator is also capable of appending new trajectories to ongoing joint motion, that is, whenever a setpoint is updated (e.g. by issuing a `positionMove` command while a previous one has not yet completed), last computed joint velocity is used to calculate the initial motion of the new trajectory.

In order to configure a control board to use this new generator, set the `TRAJECTORY_GENERATION::trajectory_type` property to `trapezoidal_speed`.

Additional changes:
* `GazeboYarpControlBoardDriver` has learned to fetch reference speed and acceleration data from initial configuration via `TRAJECTORY_GENERATION::refSpeed` and `TRAJECTORY_GENERATION::refAcceleration` properties, respectively. Existing code has been factored out and conveniently improved in a new `setTrajectoryReferences` method.
* Previous defaults have been preserved, except for now using `0.01` m/s² as max acceleration for prismatic joints.
* The control board no longer silently falls back to `minimum_jerk` if an unsupported value was passed to `TRAJECTORY_GENERATION::trajectory_type`; an error is reported instead.
* The code that initializes `trajectory_type` from config has been slightly simplified and now reports is value in a `yDebug` line.
* Added IDE-related stuff to *.gitignore* (Eclipse & VS Code).

### DIFF
--- a/plugins/controlboard/src/ControlBoardDriverTrajectory.cpp
+++ b/plugins/controlboard/src/ControlBoardDriverTrajectory.cpp
@@ -9,6 +9,7 @@
 #include <GazeboYarpPlugins/common.h>
 
 #include <ControlBoardDriverTrajectory.h>
+#include <cmath>
 #include <cstdio>
 #include <gazebo/physics/World.hh>
 #include <gazebo/physics/PhysicsEngine.hh>
@@ -107,6 +108,7 @@ TrajectoryGenerator::TrajectoryGenerator(gazebo::physics::Model* model)
 : m_robot(model)
 , m_trajectory_complete(true)
 , m_speed(0)
+, m_acceleration(0)
 , m_joint_min(0)
 , m_joint_max(0)
 {}
@@ -198,7 +200,7 @@ bool MinJerkTrajectoryGenerator::abortTrajectory (double limit)
     return ret;
 }
 
-bool MinJerkTrajectoryGenerator::initTrajectory (double current_pos, double final_pos, double speed)
+bool MinJerkTrajectoryGenerator::initTrajectory (double current_pos, double final_pos, double speed, double acceleration)
 {
     std::lock_guard<std::mutex> lock(m_mutex);
 #if GAZEBO_MAJOR_VERSION >= 8
@@ -223,6 +225,7 @@ bool MinJerkTrajectoryGenerator::initTrajectory (double current_pos, double fina
     if (m_xf > m_joint_max) m_xf = m_joint_max;
     else if(m_xf < m_joint_min) m_xf = m_joint_min;
     m_speed = speed;
+    m_acceleration = acceleration; // unused
 
     //double step = (m_trajectoryGenerationReferenceSpeed[j] / 1000.0) * m_robotRefreshPeriod * _T_controller;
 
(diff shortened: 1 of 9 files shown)


plugins/controlboard/src/ControlBoardDriverTrajectory.cpp
```diff
@@ -468,3 +471,134 @@ yarp::dev::TrajectoryType ConstSpeedTrajectoryGenerator::getTrajectoryType()
...
+        return 0;
+    }
+
+    double period = m_controllerPeriod / 1000.0; // in seconds
+    double instant = m_tick * period; // current time since start (in seconds)
+    double step = 0;
+
+    if (instant <= m_ta) // first ramp
+    {
+        step = 0.5 * m_acceleration * std::pow(period, 2.0) * (2 * m_tick + 1) + m_v0 * period;
```
traversaro, line 576: Nitpick: I think (not 100% sure to be honest) that  `std::pow(period, 2)` should use the simple period*period while specifying the exponent as double will take another more expensive overload.

plugins/controlboard/src/ControlBoardDriverTrajectory.cpp
```diff
@@ -468,3 +471,134 @@ yarp::dev::TrajectoryType ConstSpeedTrajectoryGenerator::getTrajectoryType()
...
+        m_computed_reference_velocity = m_acceleration * instant + m_v0;
+    }
+    else if (instant > m_tb) // second ramp
+    {
+        step = m_acceleration * (m_tf * period - 0.5 * std::pow(period, 2.0) * (2 * m_tick + 1));
```
traversaro, line 581: See https://github.com/robotology/gazebo-yarp-plugins/pull/527/files#r556638302 .

traversaro (changes requested):
Thanks a lot @PeterBowman ! This is a nice addition, thanks. 

I add some minor comments inline, furthermore could you update the CHANGELOG? Thanks! 

> The code that initializes trajectory_type from config has been slightly simplified and now reports is value in a yDebug line.

I would prefer actually to get rid of most of the yDebug outputs, or at least put it under an option, but indeed as you did it know for sure is consistent with the rest of the code, so I think it is ok.

traversaro:
Can you also rebase on the top of the latest master? This should fix CI.

PeterBowman:
> I add some minor comments inline, furthermore could you update the CHANGELOG? Thanks!

> Can you also rebase on the top of the latest master? This should fix CI.

Should be all done now, thanks!

traversaro:
> Edit: macOS-latest builds are still failing, but I'm not sure why.

It is probably due to actions/virtual-environments#2322, actions/virtual-environments#2391 or other GitHub Actions/Homebrew regression, feel free to ignore. Probably eventually we should add conda-based macOS CI also here, for the robotology-superbuild has proved to be much more stable then the homebrew-one.

traversaro (approved):

traversaro:
Thanks @PeterBowman !

Issue #526, opened 2021-01-11
Gazebo freezes when YARP_CLOCK is set


randaz81:
If I set the env variable:
`YARP_CLOCK=/clock`
and the I start gazebo with:
`gazebo -s libgazebo_yarp_clock.so`
then the system immediately freezes (the gazebo gui does not appear).

So, currently, the only way to work with simulated time is to have all process with YARP_CLOCK=/clock, with the only exception of gazebo that must have this env variable unset.
Thus the environment variable YARP_CLOCK cannot be put into a bash_rc file.

traversaro:
Back in time (2015/2016) this used not to be the case, but at some point it started happening, probably due to some `yarp::os::Time::delay` that is being called in some callback of the Gazebo's Physics update, that is causing a deadlock, as `yarp::os::Time::delay` will be waiting for the Gazebo time to pass to return, but the Gazebo's Physics time will not pass if the physics update does not complete because it is blocked by `yarp::os::Time::delay`. 

Is this happening in the case of empty world?

randaz81:
Yes, it freezes just with: 
`gazebo -s libgazebo_yarp_clock.so`
So the world is empty and the only plugin loaded is the clock

traversaro:
Then it is possible that there is something going on with the Network constructors or the checkNetwork method?

drdanz:
I don't remember changing anything...
A bisect is probably the best way to find where this issue was introduced

traversaro:
> 
> 
> I don't remember changing anything...
> A bisect is probably the best way to find where this issue was introduced

I am afraid that bisecting will not help as gazebo-yarp-plugins will not compile against of a YARP older then 2018/2019 . Probably it is easier to check where the program is blocked after is launched via `gazebo -s libgazebo_yarp_clock.so` .

xEnVrE:
To add some context, when launching Gazebo with that configuration on the terminal one reads

```
[INFO] |yarp.os.Port| Port /iiticublap202/gzserver/44642/clock:i active at tcp://192.168.1.5:10005/
[INFO] |yarp.os.Network| Success: port-to-port persistent connection added.
[ERROR] |yarp.os.NetworkClock| Cannot find time port "/clock" or a time topic "/clock@"
[INFO] |yarp.os.Time| Waiting for clock server to start broadcasting data ...
[INFO] |yarp.os.Time| Waiting for clock server to start broadcasting data ...
[...]

Issue #536, opened 2021-02-11
ControlBoardControlTest fails in a non deterministic way


traversaro:
This can be easily reproduced locally with:
~~~
ctest -R "Control" --output-on-failure --repeat-until-fail 1000
~~~

traversaro:
Failure are of two kind:

## ControlBoard mode not set 
~~~
[INFO] |yarp.dev.PolyDriver| Created device <remote_controlboard>. See C++ class RemoteControlBoard for documentation.
/home/traversaro/robotology-superbuild/src/GazeboYARPPlugins/tests/ControlBoardControlTest.cc:104: Failure
Expected equality of these values:
  controlMode
    Which is: 1634105192
  VOCAB_CM_POSITION
    Which is: 7565168
[Dbg] [ControlBoardControlTest.cc:110] Setting desired position via YARP to 2
/home/traversaro/robotology-superbuild/src/GazeboYARPPlugins/tests/ControlBoardControlTest.cc:121: Failure
The difference between joint->Position(0u)*rad2deg and desiredPositionInDeg is 2, which exceeds 1e-2, where
joint->Position(0u)*rad2deg evaluates to 0,
desiredPositionInDeg evaluates to 2, and
1e-2 evaluates to 0.01.
/home/traversaro/robotology-superbuild/src/GazeboYARPPlugins/tests/ControlBoardControlTest.cc:132: Failure
Expected equality of these values:
  controlMode
    Which is: 1634105192
  VOCAB_CM_VELOCITY
    Which is: 7103862
~~~

traversaro:
Possible related issues: 
* https://github.com/osrf/gazebo/pull/2949
* https://github.com/osrf/gazebo/pull/2950

PR #537, opened 2021-02-11
Fix Gazebo freezing if YARP_CLOCK is set

traversaro:
Fix https://github.com/robotology/gazebo-yarp-plugins/issues/526 . 

TL;DR: From gazebo-yarp-plugins 3.6.0, you can again set YARP_CLOCK for the Gazebo simulation, meaning that all the threads of YARP Network Wrapper Servers will be executed with the frequency correctly synchronized with the Gazebo simulation. This also means that if you like you can also set `YARP_CLOCK`  directly in your `.bashrc` or `setup.sh` script. 

With respect to the solution discussed in https://github.com/robotology/gazebo-yarp-plugins/issues/526, i.e. : 
 
> 
> A possible solution is to switch to use the system clock after yarp::os::Network was initialized, and switch back to use the default clock once `/clock` has been created by calling `yarp::os::NetworkBase::yarpClockInit(YARP_CLOCK_DEFAULT)`. This should work fine _unless_ another Gazebo plugin called one of those methods to perturb the process-global YARP clock, but I think that is ok.

in this PR there are two modifications. 

First of all, the delayed initialization of the Clock is only done if the yarp::os::Network is not initialized, and this is checked by the `yarp::os::NetworkBase::isNetworkInitialized()` method. Secondly, calling `yarp::os::Network::useNetworkClock` does not only need the `/clock` port to exists, but also wait until something is published on that port and the NetworkClock becomes valid (see useNetworkClock). For this reason, calling `useNetworkClock` in the Gazebo Physics thread is going to deadlock, as the Gazebo physics thread is also the one that is publish the new time on the clock. To avoid this, the `yarp::os::Network::useNetworkClock`  is called as part of new `std::thread` that is just created for this.

The PR also contains a regression test for the issue, that was failing before the fix. As the issue was quite related to the inner working of YARP, it was not use `NetworkBase::setLocalMode` as in the robotinterface and controlboard tests, but rather yarpserver is actually launched before the tests by using [`tiny-process-library`](https://gitlab.com/eidheim/tiny-process-library).

### DIFF
--- a/plugins/clock/src/Clock.cc
+++ b/plugins/clock/src/Clock.cc
@@ -68,7 +68,20 @@ namespace gazebo
 
     void GazeboYarpClock::Load(int _argc, char **_argv)
     {
-        m_network = new yarp::os::Network();
+        // To avoid deadlock during initialation if YARP_CLOCK is set,
+        // if the YARP network is not initialized we always initialize
+        // it with system clock, and then we
+        // This avoid the problems discussed in https://github.com/robotology/gazebo-yarp-plugins/issues/526
+        bool networkIsNotInitialized = !yarp::os::NetworkBase::isNetworkInitialized();
+
+        if (networkIsNotInitialized) {
+            m_network = new yarp::os::Network(yarp::os::YARP_CLOCK_SYSTEM);
+            m_resetYARPClockAfterPortCreation = true;
+        } else {
+            m_network = new yarp::os::Network();
+            m_resetYARPClockAfterPortCreation = false;
+        }
+
         if (!m_network
             || !yarp::os::Network::checkNetwork(GazeboYarpPlugins::yarpNetworkInitializationTimeout)) {
             yError() << "GazeboYarpClock::Load error: yarp network does not seem to be available, is the yarpserver running?";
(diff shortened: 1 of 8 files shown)


plugins/clock/src/Clock.cc
```diff
@@ -68,7 +68,20 @@ namespace gazebo
 
     void GazeboYarpClock::Load(int _argc, char **_argv)
     {
-        m_network = new yarp::os::Network();
+        // To avoid deadlock during initialation if YARP_CLOCK is set,
```
xEnVrE, line 71: Suggested change:
initialation -> initialization

xEnVrE (approved):
LGTM :+1: 

I only added few comments regarding a typo and a potential missing sentence in the code comments

plugins/clock/src/Clock.cc
traversaro: Thanks, feel free to do this suggestions with GitHub's integrated system!

traversaro, line 71:
```suggestion
        // To avoid deadlock during initializationif YARP_CLOCK is set,
```
Like this!

plugins/clock/src/Clock.cc
```diff
@@ -68,7 +68,20 @@ namespace gazebo
...
+        // if the YARP network is not initialized we always initialize
+        // it with system clock, and then we
```
traversaro, line 73:
```suggestion
        // it with system clock, and then we switch back to the default clock later
```

traversaro:
> 
> 
> LGTM 👍
> 
> I only added few comments regarding a typo and a potential missing sentence in the code comments

I think I have addressed them. The tests are failing for the same reason mentioned in https://github.com/robotology/gazebo-yarp-plugins/issues/536, so as the tests will be suppressed in https://github.com/robotology/gazebo-yarp-plugins/pull/532 I think we can merge if you agree.

xEnVrE:
>  I think we can merge if you agree.

Fine with me!

PR #541, opened 2021-02-24
[modelposepublisher] Fix broken link in documentation

xEnVrE:
After #539 and after taking a look at the documentation, I realized that the documentation of plugin `modelposepublisher` contains a link pointing to a Gazebo issue stored in the old bitbucket codebase (and now not accessible since that repository on bitbucket "does not have issue tracking enabled").

I checked on GitHub and the issue is still open. Hence, this PR updates the link to the issue.


traversaro (approved):

traversaro:
Thanks!

Issue #543, opened 2021-03-03
libYARP_robotinterface not working correctly with sensor plugins


traversaro:
Reported by @drdanz . 

The logic in https://github.com/robotology/gazebo-yarp-plugins/blob/master/libraries/singleton/src/Handler.cc#L257 assumes that the sensorScopedName is prefixed by the modelScopedName. Apparently this is not true, and instead the sensor scoped name has an additional `default::` at the beginning.

PR #544, opened 2021-03-03
Fix for getDevicesAsPolyDriverList method used by libYARP_robotinterface

traversaro:
In particular, ensure that also devices registered by sensor plugins are correctly exposed to libYARP_robotinterface.

Fix https://github.com/robotology/gazebo-yarp-plugins/issues/543 .

cc @drdanz @randaz81

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
 
 The format of this document is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Fixed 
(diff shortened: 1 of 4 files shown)


drdanz (approved):
LGTM, tested with ROS and ROS2 nws

traversaro:
Actually `default` it just happens to be the world name, we should get the world name and use that instead of hardcoding `default::`, looking into it.

traversaro:
@drdanz I updated the code to be generic w.r.t. to world name, could you check if this is still working for you? Thanks!

drdanz (commented):
Tested again, it works, thanks

traversaro:
Thanks @drdanz !

PR #550, opened 2021-05-13
Debug Gazebo issue 3001

traversaro:
Do not merge, this is just to trigger a CI job to help debugging https://github.com/osrf/gazebo/issues/3001 .

PR #551, opened 2021-05-13
Check if Gazebo 3001 issue is solved

traversaro:
Differently from https://github.com/robotology/gazebo-yarp-plugins/pull/550, this PR does not change anything in the CI scripts.


traversaro:
Problem solved: osrf/gazebo#3001 .

PR #558, opened 2021-06-28
yarp::dev::IRgbVisualParams in multicamera and depthCamera

drdanz:
Unfortunately, at the moment, in YARP we don't have an interface for a multicamera. This means that, if the camera parameters are different, this will work properly only for the first camera.

Depends on #560

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ## [Unreleased]
 
+### Added
+- The multicamera plugin now implements the `yarp::dev::IRgbVisualParams` interface.
+
+### Fixed
+- Fixed the getRgbIntrinsicParam method in the depthCamera plugin when the distortion is not set.
+- The property returned by `getRgbIntrinsicParam()`, now contains `rectificationMatrix` instead of `rectificationMatrix` (See also https://github.com/robotology/yarp/pull/2593).
+
 ## [3.6.1] - 2021-05-19
 
 ### Fixed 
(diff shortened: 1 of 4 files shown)


traversaro:
Just to clarify, the source of this code is the existing code in `gazebo_yarp_depthcamera`, right? Not sure if you want to fix the `retificationMatrix` typo after https://github.com/robotology/yarp/pull/2593 ?

traversaro:
Can you update the CHANGELOG?

traversaro:
To fix the CI, you may want to rebase over latest devel to get https://github.com/robotology/gazebo-yarp-plugins/pull/557 .

drdanz:
> Just to clarify, the source of this code is the existing code in gazebo_yarp_depthcamera, right? Not sure if you want to fix the retificationMatrix typo after robotology/yarp#2593 ?

Yes, I was saving it for the next PR, where I was fixing this, and the depthcamera, and also fixing getRgbIntrinsicParam when the distortion is not set, but perhaps it is better to have it all in one PR, I'll update it...

traversaro:
> 
> 
> > Just to clarify, the source of this code is the existing code in gazebo_yarp_depthcamera, right? Not sure if you want to fix the retificationMatrix typo after [robotology/yarp#2593](https://github.com/robotology/yarp/pull/2593) ?
> 
> Yes, I was saving it for the next PR, where I was fixing this, and the depthcamera, and also fixing getRgbIntrinsicParam when the distortion is not set, but perhaps it is better to have it all in one PR, I'll update it...

Ok, let's leave it for the next PR then!

drdanz:
> Ok, let's leave it for the next PR then!

Too late, just updated the PR 😆

drdanz:
Should this be `master` for gazebo-yarp-plugin` `devel` branch?
https://github.com/robotology/gazebo-yarp-plugins/blob/90df1566fc05e5830252bbfc1d34df4b7a4ca976/.github/workflows/ci.yml#L31

drdanz:
Rebased over #560 in order to build with YARP master

drdanz:
@traversaro: All checks have passed. Can we merge this, so that I can rebase #559?

traversaro (approved):

drdanz:
Thanks @traversaro

PR #559, opened 2021-06-28
Handle `yarpDeviceName` in controlboard and multicamera plugins

drdanz:
Depends on #558 (not really, just the changelog, to avoid merge conflicts)

### DIFF
--- a/plugins/controlboard/src/ControlBoard.cc
+++ b/plugins/controlboard/src/ControlBoard.cc
@@ -159,7 +159,15 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
                 virt_group.append(yarp::os::Bottle(net));                    
             }
             
-            std::string scopedDeviceName = m_robotName + "::" + newPoly.key.c_str();
+            std::string scopedDeviceName;
+            if(!m_parameters.check("yarpDeviceName"))
+            {
+                scopedDeviceName = m_robotName + "::" + newPoly.key.c_str();
+            }
+            else
+            {
+                scopedDeviceName = m_robotName + "::" + m_parameters.find("yarpDeviceName").asString();
+            }
             newPoly.poly = GazeboYarpPlugins::Handler::getHandler()->getDevice(scopedDeviceName);
             if( newPoly.poly != NULL)
             {
@@ -200,23 +208,13 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
             }
 
             //Register the device with the given name
-            if(!m_parameters.check("yarpDeviceName"))
+            if(!GazeboYarpPlugins::Handler::getHandler()->setDevice(scopedDeviceName, newPoly.poly))
             {
-                yError()<<"GazeboYarpControlBoard: cannot find yarpDeviceName parameter in ini file.";
+                yError() << "GazeboYarpControlBoard: failed setting scopedDeviceName(=" << scopedDeviceName << ")";
+                return;
             }
-            else
-            {
-                std::string robotName = _parent->GetScopedName();
-                std::string deviceId = m_parameters.find("yarpDeviceName").asString();
-                std::string scopedDeviceName = robotName + "::" + deviceId;
+            yInfo() << "GazeboYarpControlBoard: Registered YARP device with instance name:" << scopedDeviceName;
 
-                if(!GazeboYarpPlugins::Handler::getHandler()->setDevice(scopedDeviceName, newPoly.poly))
-                {
-                    yError()<<"GazeboYarpControlBoard: failed setting scopedDeviceName(=" << scopedDeviceName << ")";
-                    return;
-                }
-                yInfo() << "GazeboYarpControlBoard: Registered YARP device with instance name:" << scopedDeviceName;
-            }
             m_controlBoards.push(newPoly);
         }
         
@@ -259,4 +257,4 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
             return;
         }
     }
-}
+} // namespace gazebo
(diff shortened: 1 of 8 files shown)


plugins/depthCamera/src/DepthCamera.cc
```diff
@@ -137,6 +137,7 @@ void GazeboYarpDepthCamera::Load(sensors::SensorPtr _sensor, sdf::ElementPtr _sd
             yError()<<"GazeboYarpDepthCamera: failed setting scopedDeviceName(=" << scopedDeviceName << ")";
             return;
         }
+        yDebug() << "GazeboYarpDepthCamera: register device:" << scopedDeviceName;
```
traversaro, line 140: Is this strictly necessary? The device output of the gazebo yarp plugins is already quite verbose, if this was used for debugging I think we can remove it now.

traversaro:
Could you update the docs in https://github.com/robotology/gazebo-yarp-plugins/blob/40ef333672b9cc85cdf50255d229b6bb462237cb/plugins/robotinterface/README.md#how-to-specify-existing-yarp-devices-to-which-to-attach with the changes w.r.t. the use of `yarpDeviceName`?

plugins/depthCamera/src/DepthCamera.cc
drdanz: I think it is quite important to be able to search in the log for the name of the registered device (which is why I enabled it for the depthCamera and the doubleLaser). When creating a yarprobotinterface xml file, you have no other clue, unless this line is printed...

traversaro: If this is an important information to print, can we put this as `yInfo` instead?

traversaro: Furthermore, it may not be clear to user what "device" refers to, perhaps we could be a bit more explicit and:
```suggestion
        yDebug() << "GazeboYarpDepthCamera: register YARP device with instance name:" << scopedDeviceName;
```

plugins/controlboard/src/ControlBoard.cc
```diff
@@ -198,7 +198,25 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
...
+                std::string deviceId = m_parameters.find("yarpDeviceName").asString();
+                std::string scopedDeviceName = robotName + "::" + deviceId;
+
+                if(!GazeboYarpPlugins::Handler::getHandler()->setDevice(scopedDeviceName, newPoly.poly))
+                {
+                    yError()<<"GazeboYarpControlBoard: failed setting scopedDeviceName(=" << scopedDeviceName << ")";
+                    return;
+                }
+                yInfo() << "GazeboYarpControlBoard: Registered YARP device with instance name:" << scopedDeviceName;
+            }
```
traversaro, line 219: Sorry, before reading the documentation I did not understood that this was breaking the previous pattern of specifying the YARP instance name with `networks`. This would break all the existing non-`gazebo_yarp_robotinterface` use of the device map contained in the `GazeboYarpPlugins::Handler`. In particular, this would break all the existing `gazebo_yarp_controlboard` plugins that access existing devices via `gazebo_yarp_controlboard` itself, see for example https://github.com/robotology/icub-models/blob/1c9dd81f5db605f00bdc26ddec5af614e41c98b5/iCub/conf_icub3/gazebo_icub_left_hand.ini or similar configuration files. Unfortunatly this is not really feasible, unless we first update all such uses in our repo and we clearly document this breaking change for external users. I think that keeping the old behaviour if `yarpDeviceName` is not specified is probably an acceptable compromise.

traversaro (changes requested):

plugins/controlboard/src/ControlBoard.cc
drdanz: I don't think that is breaking anything... I updated the documentation according to https://github.com/robotology/yarp/pull/2482 since you can pass a "device" param or a "networks" paramlist to the attach action... The name to be used is the same.
The network linked here, is a different network i.e. the list of devices expected by the driver... when the attach action is executed, this list is matched with the names in the PolyDriverList, which is generated from the paramlist in the action.

traversaro: Unless I am wrong, now if `yarpDeviceName` is not set, `GazeboYarpPlugins::Handler::getHandler()->setDevice` is not called. In existing models such as `iCubGazeboV2_5_visuomanip` in icub-models, this means that for example for the `left_hand_finger` device in https://github.com/robotology/icub-models/blob/1c9dd81f5db605f00bdc26ddec5af614e41c98b5/iCub/conf/gazebo_icub_left_hand_thumb.ini#L21, `GazeboYarpPlugins::Handler::getHandler()->setDevice`  is not called anymore. This means that then, the wrapper with port prefix `/icubSim/left_hand` instantiated by https://github.com/robotology/icub-models/blob/1c9dd81f5db605f00bdc26ddec5af614e41c98b5/iCub/conf_icub3/gazebo_icub_left_hand.ini is not able to find and attach to the `left_hand_finger` device in https://github.com/robotology/gazebo-yarp-plugins/blob/40ef333672b9cc85cdf50255d229b6bb462237cb/plugins/controlboard/src/ControlBoard.cc#L163 , meaning that the ports starting with `/icubSim/left_hand` will not correctly stream and control the joints corresponding to `left_hand_finger`  .

drdanz: This was for coherence with the other devices using `yarpDeviceName`... Probably this is wrong everywhere, if the option is not set, it should just use the old naming style...
If you agree, I will update all the devices...

drdanz: I changed the logic in all plugins... They should now be registering with the default name if `yarpDeviceName` is not set

traversaro: Not sure of the implications in changing that on all plugins instead of just in `gazebo_yarp_controlboard` that was the only one using this functionality between `gazebo_yarp_robotinterface`, however as long as the documentation is updated to match the actual behaviour of the plugin I am ok with this (keeping the change just in  `gazebo_yarp_controlboard` permitted to easily document what was happening).

traversaro:
Hi @drdanz, thanks for the edit in the docs, can I further modify it?

drdanz:
Thanks @traversaro

PR #560, opened 2021-06-28
CI: Build using yarp master

drdanz:


traversaro:
I think we can actually remove it altogether.

.github/workflows/ci.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           - os: ubuntu-18.04
             combination: 1
             gazebo_version: gazebo10
-            yarp_version: yarp-3.4
+            yarp_version: master
```
traversaro, line 32:
```suggestion
```

traversaro: I think we can simply remove it, otherwise we need to change it back to the version specific branch once we merge devel in master.

PR #564, opened 2021-07-27
Added log component

ste93:
Component to yarp log has been added in:
- LaserSensor
- DoubleLaser
- ControBoard
- DepthCamera
- MultiCamera


plugins/controlboard/src/ControlBoardDriverPidControl.cpp
```diff
@@ -11,20 +11,22 @@
 #include <yarp/os/LogStream.h>
 #include <gazebo/physics/Joint.hh>
 #include <gazebo/transport/Publisher.hh>
-
+namespace {
+    YARP_LOG_COMPONENT(GAZEBOCONTROLBOARD, "gazebo-yarp-plugins.plugins.GazeboYarpControlBoard")
```
traversaro, line 15: This `YARP_LOG_COMPONENT` macro is duplicated among several different .cpp files, is that normal/expected?

ste93: Actually it was a mistake. But there are 2 options:
- if we want to log also the file we can keep them there and add the class to component
- otherwise I can put it in ControlBoard.hh file

traversaro:
I am a bit confused on the use of non-namespaced names as the first argument of `YARP_LOG_COMPONENT` macros. That macro essential defines a function in the global namespace (see https://github.com/robotology/yarp/blob/ea6ed180ddb102679d8b8492fc40329719d7b8e4/src/libYARP_os/src/yarp/os/LogComponent.h#L78), and defining functions with non-namespaced names in the global namespace seems a bit dangerous. @drdanz what's your opinion on this?

plugins/controlboard/src/ControlBoardDriverPidControl.cpp
traversaro: I think keeping a mapping between a class and a component is ok, no need for further granularity. However, from what I understand `YARP_LOG_COMPONENT` defines  a function with a static member, so you can't put it in a header, as this will result in ODR (one definition rule)-violations.

ste93:
I don't know if it is the right way, but we can wait for Daniele. Otherwise I can put them in the local namespace and keep them inside the file.

plugins/controlboard/src/ControlBoardDriverPidControl.cpp
ste93: Actually I tried to put it inside the h file and it gaves me some undeclared error, so I think that sould be kept inside every file

traversaro: If you see https://github.com/robotology/yarp/blob/ea6ed180ddb102679d8b8492fc40329719d7b8e4/src/libYARP_os/src/yarp/os/LogComponent.h#L78, you see that `YARP_LOG_COMPONENT` **defines** a function, while `YARP_DECLARE_LOG_COMPONENT` declares it. So I think you should do the same thing that you do if you are creating a function: you don't need to copy&paste it in every file, but rather declare it in the header, and then placing the implementaton in only a single cpp file.

drdanz:
When a component is used only in a cpp file, defining it inside an anonymous namespace is in my opinion the best option.
If you want to use it in several files, then you have use `YARP_DECLARE_LOG_COMPONENT` in a .h file, and therefore you cannot enclose it in an anonymous namespace.
I have been using this in a few plugins, and also in impl headers `yarp::os`, but since the default visibility for symbols in yarp is `hidden` and symbols must be exported with `YARP_xxx_API`, they are available when linking the library, but are not exported outside of the library. For example:

```
$ nm -ACD --defined-only lib/yarp/yarp_portmonitor.so

lib/yarp/yarp_portmonitor.so:000000000001e18c W std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > std::operator+<char, std::char_traits<char>, std::allocator<char> >(char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
lib/yarp/yarp_portmonitor.so:000000000002acc8 V typeinfo for yarp::os::YarpPluginSelector
lib/yarp/yarp_portmonitor.so:00000000000214e0 V typeinfo name for yarp::os::YarpPluginSelector
lib/yarp/yarp_portmonitor.so:000000000002ac38 V vtable for yarp::os::YarpPluginSelector
lib/yarp/yarp_portmonitor.so:0000000000007ed4 T portmonitor
lib/yarp/yarp_portmonitor.so:0000000000007d35 T portmonitor_create
lib/yarp/yarp_portmonitor.so:0000000000007daa T portmonitor_destroy
lib/yarp/yarp_portmonitor.so:0000000000007e19 T portmonitor_getAbi
lib/yarp/yarp_portmonitor.so:0000000000007e79 T portmonitor_getBaseClassName
lib/yarp/yarp_portmonitor.so:0000000000007e2b T portmonitor_getClassName
lib/yarp/yarp_portmonitor.so:0000000000007e07 T portmonitor_getVersion
```

As you can see the `PORTMONITORCARRIER` function declared in the global namespace here:

https://github.com/robotology/yarp/blob/84e958661b77dc341f51811abaa81a876ec5f14b/src/carriers/portmonitor_carrier/MonitorLogComponent.cpp#L8

and

https://github.com/robotology/yarp/blob/84e958661b77dc341f51811abaa81a876ec5f14b/src/carriers/portmonitor_carrier/MonitorLogComponent.cpp#L8-L13

do not appear in the library.

I don't know if the visibility in gazebo-yarp-plugins is set to hidden or not, therefore I cannot say for sure what will be the result here.

traversaro:
> I don't know if the visibility in gazebo-yarp-plugins is set to hidden or not, therefore I cannot say for sure what will be the result here.

The default visibility is not hidden as no explicitly visibility handling machinery is present.

ste93:
Ok so I can make a component and forwarding with declare. By the way I've noticed that not all files include ControlBoard.hh, so I think make a common log.cpp class can be the right way and then import it into all the files

traversaro:
> 
> 
> > I don't know if the visibility in gazebo-yarp-plugins is set to hidden or not, therefore I cannot say for sure what will be the result here.
> 
> The default visibility is not hidden as no explicitly visibility handling machinery is present.

Given this, probably we need to namespace this declarations. @drdanz there is any way that you suggest to do that?

ste93:
I've updated with a single class for controlboard logger, @traversaro  do you think it is fine or I should put it in a namespace?

plugins/depthCamera/src/DepthCamera.cc
```diff
@@ -12,14 +12,18 @@
...
 #include <gazebo/sensors/DepthCameraSensor.hh>
 #include <yarp/dev/PolyDriver.h>
 #include <yarp/dev/IMultipleWrapper.h>
 
 
 
 GZ_REGISTER_SENSOR_PLUGIN(gazebo::GazeboYarpDepthCamera)
-
+namespace {
+    YARP_LOG_COMPONENT(GAZEBODEPTH, "gazebo-yarp-plugins.plugins.GazeboYarpDepthCamera DepthCamera.cc")
```
traversaro, line 25: This seems to follow a different style, is this intended?

plugins/multicamera/src/MultiCamera.cc
```diff
@@ -23,6 +23,9 @@
 
 GZ_REGISTER_SENSOR_PLUGIN(gazebo::GazeboYarpMultiCamera)
 
+namespace {
+    YARP_LOG_COMPONENT(GAZEBOMULTICAMERA, "gazebo-yarp-plugins.plugins.GazeboYarpMultiCamera MultiCamera.cc")
```
traversaro, line 27: Different style?

plugins/lasersensor/src/LaserSensorDriver.cpp
```diff
@@ -18,6 +18,9 @@ using namespace boost::placeholders;
 using namespace yarp::dev;
 
 const std::string YarpLaserSensorScopedName = "sensorScopedName";
+namespace {
+    YARP_LOG_COMPONENT(GAZEBOLASER, "gazebo-yarp-plugins.plugins.GazeboYarpLaserSensor")
```
traversaro, line 22: I may be wrong, but this duplicates the definition in https://github.com/robotology/gazebo-yarp-plugins/pull/564/files#diff-db6c9bf936c1f5304516f6e46161f605ba89fd45c5f9db8d8ca905bbc078f14aR24, I guess just one and only one is necessary/required.

traversaro:
> 
> 
> I've updated with a single class for controlboard logger, @traversaro do you think it is fine or I should put it in a namespace?

Given https://github.com/robotology/yarp/blob/master/src/libYARP_os/src/yarp/os/LogComponent.h#L74, each `YARP_***_LOG_COMPONENT(GAZEBOMULTICAMERA)` macro is basically defining or declaring a function called (for example), `GAZEBOMULTICAMERA`. 

So it is basically equivalent to a definition:
~~~
const yarp::os::LogComponent& GAZEBOMULTICAMERA();
~~~

As this is quite a normal functions, I was wondering:
* Why it is named all capital, if it is a normal function? This seems indeed the style followed in http://www.yarp.it/git-master/yarp_logging.html#autotoc_md814, but all capital is tipically reserver for pre-processor macros. If that's the style of YARP components no problem for me to have the same in gazebo-yarp-plugins, but I was wondering what was the rationale for that.
* As this are normal functions, there is any reason for which they should not be defined in namespace, except for the projects like YARP that do explicit visibility handling?

plugins/depthCamera/src/DepthCamera.cc
ste93: no it was a refuse, i'll check them all :)

plugins/lasersensor/src/LaserSensorDriver.cpp
ste93: I think I can put it in header only

ste93:
>     * Why it is named all capital, if it is a normal function? This seems indeed the style followed in http://www.yarp.it/git-master/yarp_logging.html#autotoc_md814, but all capital is tipically reserver for pre-processor macros. If that's the style of YARP components no problem for me to have the same in gazebo-yarp-plugins, but I was wondering what was the rationale for that.
Actually I don't know the main reason, but it is widely common in yarp. maybe @drdanz has more information on this
>     * As this are normal functions, there is any reason for which they should not be defined in namespace, except for the projects like YARP that do explicit visibility handling?
actually I don't know but I think that can be also defined in namespace

ste93:
i@traversaro I updated with last changes we agreed :)

ste93:
I'm doing some tests now and it seems that it cannot find simbols, i'm checking it

plugins/multicamera/src/MultiCameraLog.cpp
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (C) 2013-2021 Fondazione Istituto Italiano di Tecnologia RBCS & iCub Facility & ADVR
+ * Authors: see AUTHORS file.
+ * CopyPolicy: Released under the terms of the LGPLv2.1 or any later version, see LGPL.TXT or LGPL3.TXT
+ */
+
+#include "gazebo/MultiCameraLog.h"
+
+namespace GazeboYarpPlugins{
+    YARP_LOG_COMPONENT(GAZEBOMULTICAMERA, "gazebo-yarp-plugins.plugins.GazeboYarpMultiCamera MultiCamera.cc")
```
traversaro, line 10: The style of this seems different, also in the header, is this expected?

traversaro:
Thanks, I think we are almost there. Can you update the changelog in https://github.com/robotology/gazebo-yarp-plugins/blob/devel/CHANGELOG.md ? Thanks!

plugins/multicamera/src/MultiCameraLog.cpp
traversaro: cc @ste93

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 - The controlboard and multicamera plugins now handle the `yarpDeviceName` parameter (https://github.com/robotology/gazebo-yarp-plugins/pull/559).
 
 ### Changed
+- `MultiCamera`, `LaserSensor`, `DoubleLaser`, `ControlBoard` and `DepthCamera` logging now have a component and they're all yC...() style
```
traversaro, line 13: The CHANGELOG is a document that is also targeted to external users, so it is tipically better to:
* Use non-ambigous names, so for plugins refer to the plugin name that is used in SDF (see https://github.com/robotology/gazebo-yarp-plugins/blob/master/doc/embed_plugins.md#plugins-exposing-functionalities-available-on-real-robots)
* Be clear (what does mean "logging now have a component"?) 

Furthermore, tipically sentences end with a period. I tried to revise the comment considerin this, let me know if you like it. 
```suggestion
- `gazebo_yarp_multicamera`, `gazebo_yarp_lasersensor`, `gazebo_yarp_doublelaser`, `gazebo_yarp_controlboard` and `gazebo_yarp_depthCamera` plugins now log messages using the ["Log Components" YARP logging feature](http://www.yarp.it/git-master/yarp_logging.html).
```

plugins/multicamera/src/MultiCameraLog.cpp
ste93: sorry the comment was pending and it was not sent.

do you intend the include? that's because the cmake file in this class is different and doesn't include the gazebo directory by default.

CHANGELOG.md
ste93: Yes you're riight, I'm beginning to learn to do it :) I think the way you've done it is very good, thanks!

plugins/multicamera/src/MultiCameraLog.cpp
traversaro: No, I mean the `MultiCamera.cc` part in the `YARP_LOG_COMPONENT` string.

ste93: Ok I didn't noticed that, it was a refuse. I'll remove it.

traversaro:
Great! I think it is safe to merge, as we already run CI on the other jobs.

traversaro (approved):

ste93:
thanks ;)

PR #565, opened 2021-07-28
added option to remove implicit wrappers in lasersensor, controlboard, depthcamera and multicamera

ste93:
based on #564 
This PR adds a flag USE_NEW_WRAPPERS that if enable removes all the wrappers from doublelaser, lasersensor, controlboard, depthcamera and multicamera and allows the use of the new nws present in yarp 3.5.

YARP 3.5 is required now as basis

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,10 @@
 
 cmake_minimum_required(VERSION 3.5)
 PROJECT(GazeboYARPPlugins)
+option(USE_NEW_WRAPPERS "use new nws and nwc instead of old ones" OFF)
+if(USE_NEW_WRAPPERS)
+    add_definitions(-DUSE_NEW_WRAPPERS)
+endif()
 
 # Project version
 set(${PROJECT_NAME}_MAJOR_VERSION 3)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 
 cmake_minimum_required(VERSION 3.5)
 PROJECT(GazeboYARPPlugins)
-option(DISABLE_IMPLICIT_NETWORK_WRAPPERS "use new nws and nwc instead of old ones" OFF)
+option(GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS "if enabled removes default wrappers in doublelaser, lasersensor, controlboard, depthcamera and multicamera" OFF)
 
 # Project version
 set(${PROJECT_NAME}_MAJOR_VERSION 3)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,10 +4,7 @@
 
 cmake_minimum_required(VERSION 3.5)
 PROJECT(GazeboYARPPlugins)
-option(USE_NEW_WRAPPERS "use new nws and nwc instead of old ones" OFF)
-if(USE_NEW_WRAPPERS)
-    add_definitions(-DUSE_NEW_WRAPPERS)
-endif()
+option(DISABLE_IMPLICIT_NETWORK_WRAPPERS "use new nws and nwc instead of old ones" OFF)
 
 # Project version
 set(${PROJECT_NAME}_MAJOR_VERSION 3)
--- a/plugins/controlboard/include/gazebo/ControlBoard.hh
+++ b/plugins/controlboard/include/gazebo/ControlBoard.hh
@@ -49,7 +49,7 @@ public:
     void Load(physics::ModelPtr _parent, sdf::ElementPtr _sdf);
 
 private:
-    #ifndef USE_NEW_WRAPPERS
+    #ifndef GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS
     yarp::dev::PolyDriver m_wrapper;
     yarp::dev::IMultipleWrapper* m_iWrap;
     yarp::dev::PolyDriver m_virtAnalogSensorWrapper;
--- a/plugins/controlboard/include/gazebo/ControlBoard.hh
+++ b/plugins/controlboard/include/gazebo/ControlBoard.hh
@@ -13,6 +13,7 @@
 
 #include <yarp/dev/PolyDriverList.h>
 #include <yarp/dev/IVirtualAnalogSensor.h>
+
 namespace yarp {
     namespace dev {
         class IMultipleWrapper;
@@ -55,12 +56,12 @@ private:
     yarp::dev::PolyDriver m_virtAnalogSensorWrapper;
     yarp::dev::IMultipleWrapper* m_iVirtAnalogSensorWrap;
     yarp::dev::PolyDriverList m_controlBoards;
+    bool m_useVirtAnalogSensor = false;
     #else
     yarp::dev::PolyDriver m_controlboardDriver;
     std::string m_scopedDeviceName;
     std::string m_yarpDeviceName;
     #endif
-    bool m_useVirtAnalogSensor = false;
 
     yarp::os::Property m_parameters;
 
--- a/plugins/controlboard/include/gazebo/ControlBoard.hh
+++ b/plugins/controlboard/include/gazebo/ControlBoard.hh
@@ -13,7 +13,6 @@
 
 #include <yarp/dev/PolyDriverList.h>
 #include <yarp/dev/IVirtualAnalogSensor.h>
-
 namespace yarp {
     namespace dev {
         class IMultipleWrapper;
@@ -50,14 +49,19 @@ public:
     void Load(physics::ModelPtr _parent, sdf::ElementPtr _sdf);
 
 private:
+    #ifndef USE_NEW_WRAPPERS
     yarp::dev::PolyDriver m_wrapper;
     yarp::dev::IMultipleWrapper* m_iWrap;
-    yarp::dev::PolyDriverList m_controlBoards;
-
-    bool m_useVirtAnalogSensor = false;
     yarp::dev::PolyDriver m_virtAnalogSensorWrapper;
     yarp::dev::IMultipleWrapper* m_iVirtAnalogSensorWrap;
-    
+    yarp::dev::PolyDriverList m_controlBoards;
+    #else
+    yarp::dev::PolyDriver m_controlboardDriver;
+    std::string m_scopedDeviceName;
+    std::string m_yarpDeviceName;
+    #endif
+    bool m_useVirtAnalogSensor = false;
+
     yarp::os::Property m_parameters;
 
     std::string m_robotName;
--- a/plugins/controlboard/src/ControlBoard.cc
+++ b/plugins/controlboard/src/ControlBoard.cc
@@ -23,7 +23,7 @@ namespace gazebo
 
 GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
 
-    #ifndef USE_NEW_WRAPPERS
+    #ifndef GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS
     GazeboYarpControlBoard::GazeboYarpControlBoard() : m_iWrap(nullptr),
                                                        m_iVirtAnalogSensorWrap(nullptr)
     #else
@@ -33,7 +33,7 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
 
     GazeboYarpControlBoard::~GazeboYarpControlBoard()
     {
-        #ifndef USE_NEW_WRAPPERS
+        #ifndef GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS
         if (m_iWrap) {
             m_iWrap->detachAll();
             m_iWrap = nullptr;
@@ -87,7 +87,7 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
         GazeboYarpPlugins::Handler::getHandler()->setRobot(get_pointer(_parent));
 
         // Add the gazebo_controlboard device driver to the factory.
-        #ifndef USE_NEW_WRAPPERS
+        #ifndef GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS
         yarp::dev::Drivers::factory().add(new yarp::dev::DriverCreatorOf<yarp::dev::GazeboYarpControlBoardDriver>("gazebo_controlboard", "controlboardwrapper2", "GazeboYarpControlBoardDriver"));
         #else
         yarp::dev::Drivers::factory().add(new yarp::dev::DriverCreatorOf<yarp::dev::GazeboYarpControlBoardDriver>("gazebo_controlboard", "", "GazeboYarpControlBoardDriver"));
@@ -100,7 +100,7 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
             yCError(GAZEBOCONTROLBOARD) << "File .ini not found, load failed." ;
             return;
         }
-        #ifndef USE_NEW_WRAPPERS
+        #ifndef GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS
 
         yarp::os::Bottle wrapper_group = m_parameters.findGroup("WRAPPER");
         if(wrapper_group.isNull())
--- a/plugins/controlboard/src/ControlBoard.cc
+++ b/plugins/controlboard/src/ControlBoard.cc
@@ -267,23 +267,6 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
             return;
         }
         #else
-        yarp::os::Bottle driver_group;
-        yarp::os::Bottle virt_group;
-
-        m_useVirtAnalogSensor = m_parameters.check("useVirtualAnalogSensor", yarp::os::Value(false)).asBool();
-        if (m_useVirtAnalogSensor)
-        {
-            virt_group = m_parameters.findGroup("VIRTUAL_ANALOG_SERVER");
-            if (virt_group.isNull())
-            {
-                yCError(GAZEBOCONTROLBOARD) <<"[VIRTUAL_ANALOG_SERVER] group not found in config file";
-                return;
-            }
-
-            yarp::os::Bottle& robotName_config = virt_group.addList();
-            robotName_config.addString("robotName");
-            robotName_config.addString(m_robotName.c_str());
-        }
         if(!m_parameters.check("yarpDeviceName"))
         {
             yCError(GAZEBOCONTROLBOARD) << "missing parameter yarpDeviceName";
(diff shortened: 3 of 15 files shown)


ste93:
@traversaro do you think that it should  be added a CI with the flag  USE_NEW_WRAPPERS on?

plugins/controlboard/src/ControlBoardDriverDeviceDriver.cpp
```diff
@@ -7,24 +7,23 @@
...
-
 bool GazeboYarpControlBoardDriver::open(yarp::os::Searchable& config)
 {
     m_pluginParameters.fromString(config.toString().c_str());
 
     m_deviceName = m_pluginParameters.find("name").asString().c_str();
 
     std::string robotName(m_pluginParameters.find("robotScopedName").asString().c_str());
-
+    yCDebug(GAZEBOCONTROLBOARD) << "roobotname is ========================================= " << robotName;
```
traversaro, line 23: Leftover?

ste93: yes leftover, I'll remove it

CMakeLists.txt
```diff
@@ -4,6 +4,10 @@
 
 cmake_minimum_required(VERSION 3.5)
 PROJECT(GazeboYARPPlugins)
+option(USE_NEW_WRAPPERS "use new nws and nwc instead of old ones" OFF)
```
traversaro, line 7: Using generic names it is dangerous as it can create clashes with other libraries. Could it make sense to prefix this with `GAZEBO_YARP_PLUGINS` as done with the other options defined in this project? 
```suggestion
option(GAZEBO_YARP_PLUGINS_USE_NEW_WRAPPERS "use new nws and nwc instead of old ones" OFF)
```


Furthermore, the name and description of the option feels a bit misleading to me (unless I am missing something). This option does not switch to use the new wrappers, but simply if enabled do not instantiate any network wrapper in the controlboard and sensor Gazebo plugins, leaving the task of opening the required network wrappers to the `gazebo_yarp_robotinterace` plugin. Ideally we should use a properly descriptive variable name, something like `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS`  or similar, suggestions are welcome.

CMakeLists.txt
```diff
@@ -4,6 +4,10 @@
...
+if(USE_NEW_WRAPPERS)
+    add_definitions(-DUSE_NEW_WRAPPERS)
```
traversaro, line 9: I am afraid this could lead to subtle bugs, as it is a definition that it is used in headers, even if strictly speaking this headers are not public now. Perhaps it is better to specify this as `target_compile_definitions(<target> PUBLIC <define>)`, either for the plugin affected, or if you want it to be for all targets just add it in https://github.com/robotology/gazebo-yarp-plugins/blob/master/cmake/AddGazeboYarpPluginTarget.cmake . This avoid the risk that in the future a compilatin unit links a library whose ABI that expects `USE_NEW_WRAPPERS` to be defined, and does not define it, leading to runtime crashes. If something is not clear out of this jargon, feel free to ping me on teams.

CMakeLists.txt
ste93: Yeah you're right, actually your description is correct. Both suggestion are good, I think that GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS can be good, i don't know if anyone has another suggestion, @randaz81 @drdanz

traversaro:
Thanks @ste93 ! I already commented something I noticed, even if the complete review will be easier after https://github.com/robotology/gazebo-yarp-plugins/pull/564 gets through. Could you consider reviewing the PR title to something more descriptive? "Removed wrappers"  does not sound something that really describe what this PR is doing.

ste93:
@traversaro rebased on devel

CMakeLists.txt
traversaro: Let's proceed  with this, if anyone has a better idea we can always change it later.

traversaro:
> @traversaro rebased on devel

Let me know when this is ready for review.

ste93:
> Let me know when this is ready for review.
It is ready

CMakeLists.txt
```diff
@@ -4,6 +4,7 @@
 
 cmake_minimum_required(VERSION 3.5)
 PROJECT(GazeboYARPPlugins)
+option(DISABLE_IMPLICIT_NETWORK_WRAPPERS "use new nws and nwc instead of old ones" OFF)
```
traversaro, line 7: The description of the option seems to be the old one (see discussion in https://github.com/robotology/gazebo-yarp-plugins/pull/565#discussion_r678091530) and the project-specific prefix is still missing?

plugins/controlboard/src/ControlBoard.cc
```diff
@@ -254,5 +266,67 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
...
+            if (virt_group.isNull())
+            {
+                yCError(GAZEBOCONTROLBOARD) <<"[VIRTUAL_ANALOG_SERVER] group not found in config file";
+                return;
+            }
+
+            yarp::os::Bottle& robotName_config = virt_group.addList();
+            robotName_config.addString("robotName");
+            robotName_config.addString(m_robotName.c_str());
+        }
```
traversaro, line 286: I guess that if `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` is TRUE, the `useVirtualAnalogSensor` is not relevant anymore, what do you think?

plugins/controlboard/src/ControlBoard.cc
```diff
@@ -254,5 +266,67 @@ GZ_REGISTER_MODEL_PLUGIN(GazeboYarpControlBoard)
...
+            m_parameters.put("name", m_scopedDeviceName);
+            m_parameters.put("robotScopedName", m_robotName);
+            yCDebug(GAZEBOCONTROLBOARD) << "m_parameters:"<<m_parameters.toString();
+
+            if (_sdf->HasElement("initialConfiguration")) {
+                //yCDebug(GAZEBOCONTROLBOARD)<<"Found initial Configuration: ";
+                std::string configuration_s = _sdf->Get<std::string>("initialConfiguration");
+                m_parameters.put("initialConfiguration", configuration_s.c_str());
+                //yCDebug(GAZEBOCONTROLBOARD)<<configuration_s;
+            }
```
traversaro, line 313: This code seems to be duplicate, can't it be moved outside the `#ifdef` checks?

plugins/depthCamera/src/DepthCamera.cc
```diff
@@ -123,17 +133,16 @@ void GazeboYarpDepthCamera::Load(sensors::SensorPtr _sensor, sdf::ElementPtr _sd
...
     if(!m_driverParameters.check("yarpDeviceName"))
     {
-        scopedDeviceName = m_sensorName + "::" + driver_list[0]->key;
-    }
-    else
-    {
-        scopedDeviceName = m_sensorName + "::" + m_driverParameters.find("yarpDeviceName").asString();
+        yCError(GAZEBODEPTH) << "missing yarpDeviceName parameter";
+        return;
     }
```
traversaro, line 136: This change does not seem back-compatible, and it affects the code even if `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` is OFF .

plugins/depthCamera/src/DepthCamera.cc
```diff
@@ -52,12 +52,18 @@ void GazeboYarpDepthCamera::Load(sensors::SensorPtr _sensor, sdf::ElementPtr _sd
...
 
     // Add my gazebo device driver to the factory.
+    #ifndef GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS
     ::yarp::dev::Drivers::factory().add(new ::yarp::dev::DriverCreatorOf< ::yarp::dev::GazeboYarpDepthCameraDriver>
                                         ("gazebo_depthCamera", "RGBDSensorWrapper", "GazeboYarpDepthCameraDriver"));
-
+    #else
+    ::yarp::dev::Drivers::factory().add(new ::yarp::dev::DriverCreatorOf< ::yarp::dev::GazeboYarpDepthCameraDriver>
+                                        ("gazebo_depthCamera", "", "GazeboYarpDepthCameraDriver"));
+    #endif
```
traversaro, line 61: From what I can read in http://www.yarp.it/latest/classyarp_1_1dev_1_1DriverCreatorOf.html#a0918fdc89443a9e6217bd02af78c0db6, the wrapper argument should only be empty if there is no wrapper device capable of wrapping this YARP device, but I don't think that's the case. Do you have any specific reason to put the default wrapper here and in the rest of the devices to `""`?

plugins/controlboard/src/ControlBoardDriverDeviceDriver.cpp
```diff
@@ -21,7 +21,6 @@ bool GazeboYarpControlBoardDriver::open(yarp::os::Searchable& config)
     m_deviceName = m_pluginParameters.find("name").asString().c_str();
 
     std::string robotName(m_pluginParameters.find("robotScopedName").asString().c_str());
-
```
traversaro, line 24: In general, please try to avoid touching files that are not related to the PR, as it complicates the reviewer life. If you want to cleanup a file, please do it in a separate PR or at least in a separate commit. Having said that, for this PR feel free to leave this change in, thanks!

traversaro (changes requested):

plugins/depthCamera/src/DepthCamera.cc
ste93: I left it blank because now you can have different wrappers depending on what are you doing, so there is no default one like before. Also in the new architecture we want to open the wrappers with yarprobotinterface, so I think is confusing to have it there.

plugins/controlboard/src/ControlBoard.cc
ste93: Actually I don't know very well how it works, I don't know if anyone is interested to make it relevant. I think it is not used anymore also.

plugins/depthCamera/src/DepthCamera.cc
ste93: fixed

CMakeLists.txt
ste93: fixed

plugins/depthCamera/src/DepthCamera.cc
traversaro: Ack.

plugins/controlboard/src/ControlBoard.cc
traversaro: If the `useVirtualAnalogSensor`  is on, it just creates an additional (implicit) network wrapper. Fot that reason I think we can just remove anything related to it if `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` is true.

plugins/controlboard/include/gazebo/ControlBoard.hh
```diff
@@ -50,14 +49,19 @@ public:
...
     yarp::dev::PolyDriver m_virtAnalogSensorWrapper;
     yarp::dev::IMultipleWrapper* m_iVirtAnalogSensorWrap;
-    
+    yarp::dev::PolyDriverList m_controlBoards;
+    #else
+    yarp::dev::PolyDriver m_controlboardDriver;
+    std::string m_scopedDeviceName;
+    std::string m_yarpDeviceName;
+    #endif
+    bool m_useVirtAnalogSensor = false;
```
traversaro, line 63: See related discusson on virtual analog sensor:

ste93: removed

plugins/controlboard/src/ControlBoard.cc
ste93: removed

traversaro (changes requested):
I had a leftover comment, sorry for that .

plugins/depthCamera/src/DepthCamera.cc
```diff
@@ -66,6 +72,7 @@ void GazeboYarpDepthCamera::Load(sensors::SensorPtr _sensor, sdf::ElementPtr _sd
         yCError(GAZEBODEPTH) << "error loading configuration from SDF";
         return;
     }
```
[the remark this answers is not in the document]
ste93: > I had a leftover comment, sorry for that .

no worry :) 
sorry, mi mistake. It should be outside, fixed

plugins/lasersensor/src/LaserSensor.cc
```diff
@@ -127,8 +146,11 @@ void GazeboYarpLaserSensor::Load(sensors::SensorPtr _sensor, sdf::ElementPtr _sd
     }
     else if(!driver_properties.check("yarpDeviceName"))
     {
+        #ifndef GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS
         scopedDeviceName = sensorName + "::" + driver_list[0]->key;
+        #endif
         yCError(GAZEBOLASER)<<"failed getting yarpDeviceName parameter value";
+        return;
```
traversaro, line 153: Unless I am missing something, this should only be here if `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` is defined.

traversaro:
I guess also the docs in https://github.com/robotology/gazebo-yarp-plugins/tree/master/plugins/robotinterface#how-to-specify-existing-yarp-devices-to-which-to-attach on how to specify the YARP device instance name should mention that if `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` is enabled, then the `yarpDeviceName` is used also for the `gazebo_yarp_controlboard` plugin.

traversaro:
Just one question: how are you testing that `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` set to `OFF` continue to work as intended? Are you running some kind of model that assume that the implicit wrappers are there, such as the iCub model, or the non-modified R1 model?

randaz81:
I recommend to remove plugin `doublelaser` by this PR, since it cannot work with the new architecture.
Additionally I recommend for the next release to remove it entirely from the repo since it is deprecated and no more in use.

plugins/controlboard/src/ControlBoard.cc
traversaro: Ok!

traversaro:
> I recommend to remove plugin `doublelaser` by this PR, since it cannot work with the new architecture.
> Additionally I recommend for the next release to remove it entirely from the repo since it is deprecated and no more in use.

The `gazebo_yarp_doublelaser` plugin was modified by @drdanz in https://github.com/robotology/gazebo-yarp-plugins/pull/559/files to handle the new workflow and the documentation was also updated to mention it, do you think we should remove it also from the docs?

plugins/lasersensor/src/LaserSensor.cc
traversaro: Should be this `ifndef` of `ifdef` ?

ste93:
> Just one question: how are you testing that `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` set to `OFF` continue to work as intended? Are you running some kind of model that assume that the implicit wrappers are there, such as the iCub model, or the non-modified R1 model? `gazebo-yarp-plugins` has a few tests, but unfortunatly they do not cover a lot of functionalities.

For now I'm testing with the old r1 model, I can try with iCub but I don't have any running simulator now, I'll check it maybe. Do you have any suggestion?

traversaro:
> For now I'm testing with the old r1 model, I can try with iCub but I don't have any running simulator now, I'll check it maybe.

The old R1 model seems fine, but are you testing the case in which you are not using `gazebo_yarp_robotinterface` and you are relying instead on the implicitly create wrappers, for example via yarpmotorgui?

plugins/lasersensor/src/LaserSensor.cc
ste93: ifndef since I added it later, before was without return

traversaro: Sorry, the `return` should be there only if the `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` is `ON`, so I guess that should be an `ifdef`?

traversaro: We can discuss of this on Teams if you prefer.

ste93:
> > For now I'm testing with the old r1 model, I can try with iCub but I don't have any running simulator now, I'll check it maybe.
> 
> The old R1 model seems fine, but are you testing the case in which you are not using `gazebo_yarp_robotinterface` and you are relying instead on the implicitly create wrappers, for example via yarpmotorgui?

yes I'll test with yarpmotorgui and yarpview

plugins/lasersensor/src/LaserSensor.cc
ste93: Ok, sorry I was confused. Yes it should be ifdef

ste93: I think now it is clearer

randaz81:
> The `gazebo_yarp_doublelaser` plugin was modified by @drdanz in https://github.com/robotology/gazebo-yarp-plugins/pull/559/files to handle the new workflow and the documentation was also updated to mention it, do you think we should remove it also from the docs?

Yes, with the new yarp 3.5, this device can be completely removed from the repo. 
It was existing only because it was not possible to do the attach operation which is now possible with the robotinterface plugin.
Its functionalities are now replaced by the same yarp device driver used on the real robot, i.e. https://github.com/robotology/cer/tree/master/cermod/cerDoubleLidar. This device can be now attached to a gazebo lidar device via robotinterface.

traversaro:
Thanks, the PR seems good to go. The only missing pieces are: 
* https://github.com/robotology/gazebo-yarp-plugins/pull/565#issuecomment-889717629
* Update the CHANGELOG

traversaro:
> Yes, with the new yarp 3.5, this device can be completely removed from the repo.

Cool! Can you approve PR https://github.com/robotology/gazebo-yarp-plugins/pull/570 that add a deprecation notice for the users on this, so then in the next release we can remove the plugin?

ste93:
> Thanks, the PR seems good to go. The only missing pieces are:
> 
>     * [added option to remove implicit wrappers in lasersensor, controlboard, depthcamera and multicamera #565 (comment)](https://github.com/robotology/gazebo-yarp-plugins/pull/565#issuecomment-889717629)
> 
>     * Update the CHANGELOG

updated both

ste93:
I've tested it on gazebo, yarpmotorgui, yarplaserscannergui and yarpview both with and without the new option and it seems to work fine :)

plugins/robotinterface/README.md
```diff
@@ -76,6 +76,8 @@ The plugins that spawn YARP devices in a way that they can be then attached to t
 For all the following plugins, the **YARP device instance name** can be specified by the `yarpDeviceName` parameter in the plugin configuration.
 For the `gazebo_yarp_controlboard`, if the `yarpDeviceName` parameter is not specified, for legacy reason the **YARP device instance name** for each created device can also be specified with the `networks` parameter list in the plugin configuration.
 
+A cmake option (`GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS`) has been added, if this option is enabled then implicit wrappers present in`gazebo_yarp_multicamera`, `gazebo_yarp_lasersensor`, `gazebo_yarp_controlboard` and `gazebo_yarp_depthCamera` are removed, the new way to have them is to attach the new nws to gazebo devices via yarprobotinterface as above.
```
traversaro, line 79: It is a bit strange to have a sentence that uses the present perfect in a document that uses the present, don't you find? Did you tried to check if this sentence is coherent with the rest of the document? I think it may be more clear to remove all together the new sentence, and just add a sentence to the previous period, something like:
```suggestion
If the `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` option is set to `OFF` (default value), for the `gazebo_yarp_controlboard`, if the `yarpDeviceName` parameter is not specified, for legacy reason the **YARP device instance name** for each created device can also be specified with the `networks` parameter list in the plugin configuration. If instead the `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` option is set to `ON`, the `gazebo_yarp_controlboard` behaves like the rest of the plugins and requires to take the **YARP device instance name** from the `yarpDeviceName` parameter.
```

traversaro:
> 
> 
> I've tested it on gazebo, yarpmotorgui, yarplaserscannergui and yarpview both with and without the new option and it seems to work fine :)

Great, can you check my last comment on the docs?

plugins/robotinterface/README.md
```diff
@@ -74,7 +74,10 @@ For historic reason, not all the `gazebo-yarp-plugins` support specifying the **
...
+
+If the `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` option is set to `OFF` (default value), for the `gazebo_yarp_controlboard` if the `yarpDeviceName` parameter is not specified, for legacy reason the **YARP device instance name** for each created device can also be specified with the `networks` parameter list in the plugin configuration of `yarpDeviceName`. If instead the `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` option is set to `ON`, the `gazebo_yarp_controlboard` behaves like the rest of the plugins and requires to take the **YARP device instance name** from the `yarpDeviceName` parameter. The same behaviour applies to `gazebo_yarp_depthCamera`.
+
+Instead if the `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` option is set to `ON` then the plugins `gazebo_yarp_multicamera`, `gazebo_yarp_lasersensor`, `gazebo_yarp_controlboard` and `gazebo_yarp_depthCamera` doesn't create the implicit wrapper device but is left to the user to attach it via the robotinterface.
```
traversaro, line 80: Why this sentence starts with `Instead`? Perhaps Furthermore would be a better alternative?
```suggestion
Furthermore if the `GAZEBO_YARP_PLUGINS_DISABLE_IMPLICIT_NETWORK_WRAPPERS` option is set to `ON` then the plugins `gazebo_yarp_multicamera`, `gazebo_yarp_lasersensor`, `gazebo_yarp_controlboard` and `gazebo_yarp_depthCamera` doesn't create the implicit wrapper device but is left to the user to attach it via the robotinterface.
```

traversaro (changes requested):

plugins/robotinterface/README.md
ste93: yes, it was a refuse :)

traversaro:
Thanks @ste93 ! I would Squash and merge,  but if you prefer to re-structure your commits please let me know.

ste93:
no I think you can squash the commits

traversaro (approved):

traversaro:
Thanks @ste93 !

PR #575, opened 2021-09-01
Remove support for Gazebo 9 and 10

traversaro:
See https://github.com/robotology/robotology-superbuild/issues/812 for some context.


Nicogene (approved):
I didn't go through all the files, but seems ok to me!

Issue #598, opened 2021-12-21
Compilation against YARP master failed


traversaro:
See https://github.com/robotology/robotology-superbuild/issues/959 .

traversaro:
A fix was proposed in https://github.com/robotology/gazebo-yarp-plugins/pull/597 .

traversaro:
Fixed by https://github.com/robotology/gazebo-yarp-plugins/pull/599 .

PR #604, opened 2022-01-13
updated correct timestamp GazeboYarpLaserSensorDriver

ste93:
since Lidar2DDeviceBase has a timestamp inside and lasersensor inherits from it, I would suggest use the timestamp from taht class. It has also an api getLastInputStamp ()  to get that stamp.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@ All notable changes to this project will be documented in this file.
 The format of this document is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+
 ### Fixed
-- Removed IPreciselyTimed from LaserSensorDriver and fixed bug in it (used the correct variable `m_timestamp` inherited from `yarp::dev::Lidar2DDeviceBase`) (https://github.com/robotology/gazebo-yarp-plugins/pull/604).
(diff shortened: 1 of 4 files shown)


ste93:
@randaz81

randaz81 (approved):
thank you, this was an important bugfix

randaz81:
Additionally, since yarp 3.6 has been released, I think that we can safely remove the dependency from `IPreceiselyTimed`
https://github.com/robotology/gazebo-yarp-plugins/pull/604/files#diff-a33e9fff4431518b3add5ded498eae3af09ef01c83956d6fc6abd0a70de405b8R67-R70

REASON: in general `IPreceiselyTimed interface should be avoided` as much as possible, and possibly `deprecated`. Timestamps must be obtained together with the data using a single call to a get method()**. Otherwise,  it cannot be guaranteed that the timestamp is really associated with the retrieved data. (Threads can be interrupted etc.)**

traversaro:
Yes, feel free to remove the inheritance from `IPreceiselyTimed` and require YARP 3.6 as suggested by @randaz81 , thanks!

ste93:
@randaz81 updated, for me it can be merged. FYI @traversaro

traversaro:
> @randaz81 updated, for me it can be merged. FYI @traversaro

To require YARP 3.6 you need to modify the CMake in https://github.com/robotology/gazebo-yarp-plugins/blob/master/CMakeLists.txt#L49 . Could you also update the CHANGELOG?

ste93:
related to https://github.com/robotology/gazebo-yarp-plugins/issues/598

traversaro:
Thanks @ste93 .

ste93:
thanks @traversaro

Issue #35, opened 2013-11-11
Used GAZEBO and YARP version


EnricoMingo:
I am using GAZEBO 2.0 and YARP version X.X.60/wrapper branch.

PR #608, opened 2022-01-27
Decimal places limit for depth Camera

elandini84:
The possibility to limit the decimal places of the pixels values has
been added to the DepthCameraDriver class.

The main reason behind this PR relies on the great bandwidth consumption of the depth images (32 bit float data and up to 7-8 significant digits.) even after lossless compression.

The following set of tests should provide a good explanation of the benefits of this new feature (thanks to @DatSpace for preforming the tests and for the useful table)

### Data
For the experiment we try different quantization values (significant digits) for a depth image from gazebo, while **_the robot is looking turned 45 degrees at a wall 1m away from the camera. The bandwidth bellow is per frame averaged over 10 seconds, transmitted through LAN with zlib compression_**.

| Significant Digits  | Bandwidth | Percentage Change | Comments |
| :---: | :---: | :---: | :---: |
| 8 | 7.23 Mbps | - |  |
| 7 | 7.35 Mbps | +1.66% | It appears that at that many digits it has no effect and it just seems random |
| 6 | 7.04 Mbps | -4.2% |  |
| 5 | 6.05 Mbps | -14.0% |  |
| 4 | 1.77 Mbps | -70.7% |  |
| 3 | 0.311 Mbps | -82.4% | Millimeter accuracy |
| 2 | 0.11 Mbps | -64.6% | Centimeter accuracy. This seems to be the most logical for our use case with R1 since the camera currently used (Realsense D455) has a theoretical accuracy on the z axis of <2% at 4m, which makes it to have an error of 8cm at worst, at 4m. |
| 1 | 0.099 Mbps | -9.9% | Decimeter accuracy with visible quantitation in the depth image color. |
| 0 | 0.069 Mbps | -29.7% | This represents only integer values in meters and obviously cannot be used. |

**Note:** There was some background data transfer at the range of 30kbps that is significant for the lowest measurements. Still the numbers are so small that it doesn't really matter.

NB: If the configuration parameters passed to the plugin do not contain the group `QUANT_PARAM` this feature is disabled and the plugin will work as it always had.

EDIT: An example of configuration file with the parameters group needed for this feature can be seen in [PR #35](https://github.com/robotology/cer-sim/pull/35) for the `cer-sim` repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ## [Unreleased]
 
-### Changed
-- The possibility to limit the decimal places of the pixels values has been added to the DepthCameraDriver class. The data modification is performed in the `GazeboYarpDepthCameraDriver::getDepthImage` function.
+### Added 
+- The possibility to limit the decimal places of the pixels values has been added to the `gazebo_yarp_depthcamera` plugin, via the `QUANT_PARAM::depth_quant` parameter. The data modification is performed in the `GazeboYarpDepthCameraDriver::getDepthImage` function (https://github.com/robotology/gazebo-yarp-plugins/pull/608).
 
 ### Fixed
 - Removed `getLastInputStamp` method from `LaserSensorDriver` class in `gazebo_yarp_lasersensor`. Furthermore, fix bug in `gazebo_yarp_lasersensor`, by adding the update to the variable `m_timestamp` inherited from `yarp::dev::Lidar2DDeviceBase` (https://github.com/robotology/gazebo-yarp-plugins/pull/604).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ## [Unreleased]
 
+### Changed
+- The possibility to limit the decimal places of the pixels values has been added to the DepthCameraDriver class. The data modification is performed in the `GazeboYarpDepthCameraDriver::getDepthImage` function.
+
 ### Fixed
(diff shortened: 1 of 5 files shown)


traversaro:
Hi @elandini84, sorry, I had missed this PR.

traversaro:
This is quite a cool feature! Before merging, can you:
* Update the changelog
* Add the documentation on the new parameter in https://github.com/robotology/gazebo-yarp-plugins/blob/master/plugins/depthCamera/include/gazebo/DepthCamera.hh#L47
* Add an example of the new parameter used in https://github.com/robotology/gazebo-yarp-plugins/blob/master/tutorial/model/depthcamera/depthcamera.ini ?

Thanks!

elandini84:
> This is quite a cool feature! Before merging, can you:
> 
> * Update the changelog
> * Add the documentation on the new parameter in https://github.com/robotology/gazebo-yarp-plugins/blob/master/plugins/depthCamera/include/gazebo/DepthCamera.hh#L47
> * Add an example of the new parameter used in https://github.com/robotology/gazebo-yarp-plugins/blob/master/tutorial/model/depthcamera/depthcamera.ini ?
> 
> Thanks!

I added the requested changes. Just one thig: I added the parameter inside a separate section (`[QUANT_PARAM]`). Do you think that it would be better to get rid of the config file section? If so, just tell and I'll modify the code.

traversaro:
Thanks! I modified a bit the changelog to be a bit more clear. For what regards the name of the parameter for me the choice of group/not group is indifferent, if you like it like this we can proceed.

elandini84:
> Thanks! I modified a bit the changelog to be a bit more clear. For what regards the name of the parameter for me the choice of group/not group is indifferent, if you like it like this we can proceed.

Thanks for the changes.

elandini84:
> @elandini84 so the current name is ok for you?

Yes, absolutely

PR #609, opened 2022-02-24
Fix use of yarpConfigurationString parameter in gazebo_yarp_worldinterface plugin

traversaro:
The use of `yarpConfigurationString` element to pass configuration parameters to YARP plugins via an inline string instead of an external `.ini` YARP file was discussed in https://github.com/robotology/gazebo-yarp-plugins/issues/142 has been added in:
* https://github.com/robotology/gazebo-yarp-plugins/pull/394
* https://github.com/robotology/gazebo-yarp-plugins/pull/396

However, this option was never supported in `gazebo_yarp_worldinterface` plugin. This PR ensures that even for that plugin configuration parameters can be passed via `yarpConfigurationString`.


traversaro:
Please ignore the Windows CI failure that will be fixed by https://github.com/robotology/gazebo-yarp-plugins/pull/610 .

lrapetti (approved):
Looks good, thanks

PR #620, opened 2022-04-01
gazebo_yarp_controlboard: Add coupling icub_left_hand_mk4 to model iCub mk4 left hand

mfussi66:
This PR aims to ~update the coupling rules of the mk3 hand with the new mk4 version given that the former is now obsolete~ add new coupling rules for the new iCub hand mk4, while keeping intact the mk3 version until declared obsolete.
The hand mk4 has 5 fingers, causing an increase in the number of joints. In particular, the ring and pinkie fingers move together, since they are physically actuated by one tendon only. 

The coupling laws that map the degrees of actuation into model joints are kept to the mk3 version, although they need updating since the geometric quantities have changed.

Reference documentation on the subject: 
[FingerkinV2.docx](https://github.com/robotology/gazebo-yarp-plugins/files/8408472/FingerkinV2.docx)

### DIFF
--- a/plugins/controlboard/src/ControlBoardDriver.cpp
+++ b/plugins/controlboard/src/ControlBoardDriver.cpp
@@ -327,9 +327,15 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
             }
             else if (coupling_bottle->get(0).asString()=="icub_hand_mk3")
             {
-                BaseCouplingHandler* cpl = new HandMk4CouplingHandler(m_robot,coupled_joints, coupled_joint_names, coupled_joint_limits);
+                BaseCouplingHandler* cpl = new HandMk3CouplingHandler(m_robot,coupled_joints, coupled_joint_names, coupled_joint_limits);
                 m_coupling_handler.push_back(cpl);
                 yCInfo(GAZEBOCONTROLBOARD) << "using icub_hand_mk3";
+            }
+            else if (coupling_bottle->get(0).asString()=="icub_hand_mk4")
+            {
+                BaseCouplingHandler* cpl = new HandMk4CouplingHandler(m_robot,coupled_joints, coupled_joint_names, coupled_joint_limits);
+                m_coupling_handler.push_back(cpl);
+                yCInfo(GAZEBOCONTROLBOARD) << "using icub_hand_mk4";
             }
             else if (coupling_bottle->get(0).asString()=="none")
             {
--- a/plugins/controlboard/src/ControlBoardDriver.cpp
+++ b/plugins/controlboard/src/ControlBoardDriver.cpp
@@ -175,7 +175,8 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
         m_velocity_watchdog[j] = new Watchdog(0.200); //watchdog set to 200ms
     }
 
-    yCDebug(GAZEBOCONTROLBOARD) << "done";
+    yCDebug(GAZEBOCONTROLBOARD) << "Trajectory successfully generated.";
+
     for (size_t j = 0; j < m_numberOfJoints; ++j)
     {
         m_controlMode[j] = VOCAB_CM_POSITION;
--- a/plugins/controlboard/src/ControlBoardDriver.cpp
+++ b/plugins/controlboard/src/ControlBoardDriver.cpp
@@ -327,7 +327,7 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
             }
             else if (coupling_bottle->get(0).asString()=="icub_hand_mk3")
             {
-                BaseCouplingHandler* cpl = new HandMk3CouplingHandler(m_robot,coupled_joints, coupled_joint_names, coupled_joint_limits);
+                BaseCouplingHandler* cpl = new HandMk4CouplingHandler(m_robot,coupled_joints, coupled_joint_names, coupled_joint_limits);
                 m_coupling_handler.push_back(cpl);
                 yCInfo(GAZEBOCONTROLBOARD) << "using icub_hand_mk3";
             }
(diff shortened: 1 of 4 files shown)


plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -326,7 +327,7 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
             }
             else if (coupling_bottle->get(0).asString()=="icub_hand_mk3")
             {
-                BaseCouplingHandler* cpl = new HandMk3CouplingHandler(m_robot,coupled_joints, coupled_joint_names, coupled_joint_limits);
+                BaseCouplingHandler* cpl = new HandMk4CouplingHandler(m_robot,coupled_joints, coupled_joint_names, coupled_joint_limits);
```
traversaro, line 330: This piece of code uses the `HandMk4CouplingHandler` class if the coupling name asked is `icub_hand_mk3`, I am afraid this may be confusing, what do you think? See also the subsquent message `yCInfo(GAZEBOCONTROLBOARD) << "using icub_hand_mk3";`. Perhaps we could have a specific coupling name for this?

traversaro:
Some comments on the top of the one that I did online.

Do you have a reference for the coupling laws? For the mantainability it would be great if the coupling laws are documented (i.e. written down) somewhere, and then the gazebo plugins just implement the laws documented somewhere else. This is not a strict requirement if it takes too much time (in https://github.com/robotology/gazebo-yarp-plugins/pull/469 we merged the mk3 coupling without any public reference) but it would be nice to have.

> This PR aims to update the coupling rules of the mk3 hand with the new mk4 version, given that the former is now obsolete.

I am not sure we did any announcement on that coupling being obsolete in any changelog. Furthermore, that coupling is still used in the model [`left_hand_mk3`](https://github.com/robotology/icub-models/tree/dfd56138c196542add5caf2baa5c8a730fb79a7d/iCub/robots/left_hand_mk3) in icub-models (see https://github.com/robotology/icub-models/blob/v1.23.2/iCub/conf_left_hand_mk3/left_hand_phys.ini#L15). I am totally ok on removing the `icub_hand_mk3` coupling, just I would do it in a PR different from this one, so that we can do at least a release in which we announce that we will remove  `icub_hand_mk3`.

traversaro:
By the way, I was wondering: the coupling is the same for the left and right hand, or is this a left-specific coupling?

plugins/controlboard/src/ControlBoardDriver.cpp
mfussi66: I kept the naming `icub_hand_mk3` to be back-compatible on the user-configuration level, but I agree, it might be better to give it a specific name and separate things and maybe keep both classes for the time being.

traversaro: Yes, I think in any case you would not be able to be back-compatible as the number of joints is different, right?

pattacini:
Hi @traversaro 

>Do you have a reference for the coupling laws? For the mantainability it would be great if the coupling laws are documented (i.e. written down) somewhere, and then the gazebo plugins just implement the laws documented somewhere else. This is not a strict requirement if it takes too much time (in https://github.com/robotology/gazebo-yarp-plugins/pull/469 we merged the mk3 coupling without any public reference) but it would be nice to have.

This is a good point indeed, but I'd call for someone from the Team DEV (@Andrea-Mor, @andreaderito) to address it, as we're not the original developers.

mfussi66:
> Do you have a reference for the coupling laws? For the mantainability it would be great if the coupling laws are documented (i.e. written down) somewhere, and then the gazebo plugins just implement the laws documented somewhere else. This is not a strict requirement if it takes too much time (in #469 we merged the mk3 coupling without any public reference) but it would be nice to have.

The only reference I have is the document posted in #469, I'll attach it in the PR post, at least until proper documentation on the subject is done, if it's ok.
 
> > This PR aims to update the coupling rules of the mk3 hand with the new mk4 version, given that the former is now obsolete.
> 
> I am not sure we did any announcement on that coupling being obsolete in any changelog. 

My bad! Correcting the PR post now.

> I am totally ok on removing the `icub_hand_mk3` coupling, just I would do it in a PR different from this one and also removing the `left_hand_mk3` model from icub-models, so that we can do at least a release in which we announce that we will remove `icub_hand_mk3`.
Agreed! I will add a commit to add a new HandMk4 class and keep HandMk3 for the time being.

traversaro:
> The only reference I have is the document posted in #469, I'll attach it in the PR post, at least until proper documentation on the subject is done, if it's ok.

Sure, as mention this is just a nice to have, not something that is blocking the PR.

traversaro:
@mfussi66 I updated the CHANGELOG and PR title, let me know if you think this is ok. The only remaining point is https://github.com/robotology/gazebo-yarp-plugins/pull/620#issuecomment-1087311181, do you have any idea on that? If the coupling is left-specific, probably it would be nice to already use a left-specific name for the coupling.

mfussi66:
> do you have any idea on that? If the coupling is left-specific, probably it would be nice to already use a left-specific name for the coupling.

@traversaro Honestly I did not think about that aspect, my bad. Of course the left and right hands will be mirrored, therefore some quantities' signs might change in the coupling laws. 
Maybe it's better to have a left-specific name, at this point even in the PR title and the changelog for added clarity.

traversaro:
> > do you have any idea on that? If the coupling is left-specific, probably it would be nice to already use a left-specific name for the coupling.
> 
> @traversaro Honestly I did not think about that aspect, my bad. Of course the left and right hands will be mirrored, therefore some quantities' signs might change in the coupling laws. Maybe it's better to have a left-specific name, at this point even in the PR title and the changelog for added clarity.

Thanks, yes at least for the parameter name at this point I would then use a left-specific name, for all the rest of the names (for example the class) we can use the existing one as they can always be changed later.

traversaro:
Ok for me! @mfussi66 it is ok for you if I merge with squash to get a clean history?

mfussi66:
> Ok for me! @mfussi66 it is ok for you if I merge with squash to get a clean history?

@traversaro even better!

PR #621, opened 2022-04-04
Merge devel in master

traversaro:


traversaro:
No need for review as all the changes in devel were reviewed when inserted in devel.

Issue #384, opened 2018-09-12
Migrate gazebo_yarp_forcetorque to use the  new YARP Multiple Analog Sensor Interfaces 


traversaro:
See http://www.yarp.it/group__dev__iface__multiple__analog.html and https://github.com/robotology/yarp/issues/1526 .

traversaro:
It should be possible to support both the new and old wrappers.

traversaro:
To clarify, I think that as done for the `gazebo_yarp_imu` in https://github.com/robotology/gazebo-yarp-plugins/pull/443, this should be done in a way that it would be possible to stream data from gazebo_yarp_forcetorque  plugin using a  multipleanalogsensorsserver device.

PR #628, opened 2022-06-19
Expose yarp::dev::ISixAxisForceTorqueSensors interface in gazebo_yarp_forcetorque plugin

traversaro:
Fix https://github.com/robotology/gazebo-yarp-plugins/issues/384 . 

This is important to correctly simulate the iCub 3 model, that as of June 2022 exposes the FT sensors only via ports opened by the `multipleanalogsensorsserver`.

### DIFF
--- a/plugins/forcetorque/src/ForceTorqueDriver.cpp
+++ b/plugins/forcetorque/src/ForceTorqueDriver.cpp
@@ -150,7 +150,7 @@ size_t GazeboYarpForceTorqueDriver::getNrOfSixAxisForceTorqueSensors() const
 
 yarp::dev::MAS_status GazeboYarpForceTorqueDriver::getSixAxisForceTorqueSensorStatus(size_t sens_index) const
 {
-    if (sens_index > 1)
+    if (sens_index >= 1)
     {
         return MAS_UNKNOWN;
     }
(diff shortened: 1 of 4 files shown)


plugins/forcetorque/src/ForceTorqueDriver.cpp
```diff
@@ -139,6 +142,67 @@ int GazeboYarpForceTorqueDriver::calibrateChannel(int /*ch*/, double /*v*/)
...
 
+// SIX AXIS FORCE TORQUE SENSORS
+size_t GazeboYarpForceTorqueDriver::getNrOfSixAxisForceTorqueSensors() const
+{
+    return 1;
+}
+
+yarp::dev::MAS_status GazeboYarpForceTorqueDriver::getSixAxisForceTorqueSensorStatus(size_t sens_index) const
+{
+    if (sens_index > 1)
```
HosameldinMohamed, line 153: Hi @traversaro 

Just to double-check about this condition.. I checked about `sens_index`, and in https://www.yarp.it/v3.7/classyarp_1_1dev_1_1ISixAxisForceTorqueSensors.html it says

```
[in] | sens_index | The index of the specified sensor (should be between 0 and getNrOfSixAxisForceTorqueSensors()-1).
```

So in our case only `0` is acceptable if I am not mistaken.

traversaro: Yes, good catch!

traversaro: Fixed in https://github.com/robotology/gazebo-yarp-plugins/pull/628/commits/0662a7a77de0c1baa5e39f8d7957b05314bb9033 .

HosameldinMohamed (approved):

PR #633, opened 2022-08-26
Fix behaviour of MultipleAnalogSensor get**Name methods in nested models

traversaro:
The methods such as:
* `getThreeAxisGyroscopeName`
* `getThreeAxisGyroscopeFrameName`
* `getSixAxisForceTorqueSensorName`
* `getSixAxisForceTorqueSensorFrameName`

and similar where not working correctly in nested models. In particular, they were reporting the partially scoped sensor name. For example, in the model "iCubGenova09 fixed" in https://github.com/robotology/icub-models-generator/pull/222 that was the one on which I found the issue, the sensor name:
~~~
l_foot_rear_ft_sensor
~~~
was reported as:
~~~
iCub::l_foot_rear_ft_sensor
~~~

This PR fixes the problem by always returning the part of the sensor name after the `::` part.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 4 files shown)


mebbaid (approved):
lgtm

PR #636, opened 2022-09-20
[DO NOT MERGE] Reproduce macOS/Conda with vtk regression experienced in https://github.com/robotology/robotology-superbuild/pull/1217

traversaro:
See https://github.com/robotology/robotology-superbuild/pull/1217 .


traversaro:
> See [robotology/robotology-superbuild#1217](https://github.com/robotology/robotology-superbuild/pull/1217) .

That problem is not appearing. -_-

Issue #592, opened 2021-12-03
Wrong data streamed by the accelerometer on the iCub3 head


isorrentino:
Here is what I see logging data received on the yarp port `/icubSim/head/inertials`.

![Screenshot from 2021-12-03 17-36-09](https://user-images.githubusercontent.com/43743081/144639046-a161c159-e4cb-452c-8a66-ec60ad20b6ae.png)

I expect to see the gravity in the `z` direction instead of $10^{-4}$.

I tried with two models: `iCubGazeboV3_fixed (no hands)` and `iCubGazeboV3 (no hands)`.

cc @prashanthr05 @traversaro

traversaro:
Can you post a few lines of the dump of `/icubSim/head/inertials` ?

isorrentino:
I dumped data by using the `YarpRobotLoggerDevice` in `BipedalLocomotionFramework` (https://github.com/ami-iit/bipedal-locomotion-framework/tree/master/devices/YarpRobotLoggerDevice).
Here is the .mat file:
[yarprobot_dataset_2021_12_03_17_34_55.zip](https://github.com/robotology/gazebo-yarp-plugins/files/7651363/yarprobot_dataset_2021_12_03_17_34_55.zip)

The `mas remapper` is defined in https://github.com/ami-iit/bipedal-locomotion-framework/blob/master/devices/YarpRobotLoggerDevice/app/robots/iCubGazeboV3/interface/mas-remapper.xml and the `multipleanalogsensorsclient` is https://github.com/ami-iit/bipedal-locomotion-framework/blob/master/devices/YarpRobotLoggerDevice/app/robots/iCubGazeboV3/interface/head_imu.xml

traversaro:
Can you read/dump at least a few lines of the port that starts with `/icubSim/head/inertials` with `yarp read`?  So that we can at least understand at which point of the stack the problem is.

isorrentino:
Here is a dump of `yarp read ... /icubSim/head/inertials/measures:o`

[port.txt](https://github.com/robotology/gazebo-yarp-plugins/files/7652543/port.txt)

traversaro:
fyi @gabrielenava that also reported strage results from a `gazebo_yarp_imu` plugin recently.

gabrielenava:
@FabioBergonti suggested me to have a look at https://github.com/ami-iit/element_floating-base-estimation/issues/61#issuecomment-651695040

[...]

PR #639, opened 2022-11-04
Reference initial orientation to world frame in IMU plugin

PeterBowman:
**Problem:**
The IMU on our robot is mounted in such a way that the sensor frame does not match the inertial/world frame. Adjustments can be made in the SDF model file so that simulated angular velocities and linear accelerations are expressed in the local sensor frame, as expected. However, orientation measurements will always be set to zero by Gazebo at initialization, whereas the real IMU accounts for the frame mismatch and actually returns non-zero values (according to the influence of gravity).

**Proposed solution:**
The sensor can be configured to measure orientations in terms of the world frame via `ImuSensor::SetWorldToReferenceOrientation`. I adapted the solution from the ros-simulation repo linked below. A new configuration option, "useWorldReferenceOrientation", enables this opt-in behavior.

As a side note, we are using an Xsens MTi-28A53G35 sensor, wrapped by the "xsensmtx" YARP device located in [icub-main](https://github.com/robotology/icub-main/tree/9ecb2a8a6c8c7c15cd6b7e626c58443ac6589aea/src/libraries/icubmod/xsensmtx).

See also:
* https://github.com/gazebosim/gazebo-classic/issues/1959
* https://github.com/ros-simulation/gazebo_ros_pkgs/pull/1051
* https://github.com/robotology/gazebo-yarp-plugins/issues/592#issuecomment-1115892822

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,9 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ## [Unreleased]
 
-### Added
+### Changed
 
-- Added a new `useWorldReferenceOrientation` config option in the `gazebo_imu` plugin to take account for any initial non-zero orientation of the sensor as measured via `yarp::dev::IOrientationSensors` (https://github.com/robotology/gazebo-yarp-plugins/pull/639).
+- The `gazebo_imu` plugin used to output orientation measurements as if the sensor was zero-aligned with the world frame, regardless of the initial orientation of the part the sensor is attached to. To match most real-world IMUs, the default behavior was changed to allow measuring orientation with regard to the world frame at all times. The old behavior can be restored by setting the `useInitialSensorOrientationAsReference` config option (https://github.com/robotology/gazebo-yarp-plugins/pull/639).
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,11 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ## [Unreleased]
 
-### Fixed 
+### Added
+
+- Added a new `useWorldReferenceOrientation` config option in the `gazebo_imu` plugin to take account for any initial non-zero orientation of the sensor as measured via `yarp::dev::IOrientationSensors` (https://github.com/robotology/gazebo-yarp-plugins/pull/639).
+
+### Fixed
 
 - Fix wrong install include for gazebo_yarp_lib_common library (https://github.com/robotology/gazebo-yarp-plugins/pull/644).
 
--- a/plugins/imu/src/IMU.cc
+++ b/plugins/imu/src/IMU.cc
@@ -152,8 +152,8 @@ void GazeboYarpIMU::Load(sensors::SensorPtr _sensor, sdf::ElementPtr _sdf)
                                           {"sensor_name", ::yarp::os::Value{_sensor->Name()}}
                                         };
 
-    if (m_parameters.check("useWorldReferenceOrientation")) {
-        imu_properties.put("useWorldReferenceOrientation", yarp::os::Value::getNullValue());
+    if (m_parameters.check("useInitialSensorOrientationAsReference")) {
+        imu_properties.put("useInitialSensorOrientationAsReference", yarp::os::Value::getNullValue());
     }
 
     //Open the driver
--- a/plugins/imu/src/IMU.cc
+++ b/plugins/imu/src/IMU.cc
@@ -152,6 +152,9 @@ void GazeboYarpIMU::Load(sensors::SensorPtr _sensor, sdf::ElementPtr _sdf)
                                           {"sensor_name", ::yarp::os::Value{_sensor->Name()}}
                                         };
 
+    if (m_parameters.check("useWorldReferenceOrientation")) {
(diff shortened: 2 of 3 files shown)


traversaro:
Sorry for the late review @PeterBowman ! 

> The IMU on our robot is mounted in such a way that the sensor frame does not match the inertial/world frame.

Just for me to understand, in this context you mean that th sensor frame *in the initial pose* does not match the inertial/world frame, right? Because in general I guess the sensor frame it can't always be aligned with the world frame, as it it moves attached to the link to the sensor is attached.

PeterBowman:
Hi, @traversaro! IMU data (generated by the Gazebo-YARP plugin) is inconditionally offset from the world frame, i.e. the sensor frame is **initially aligned** with it regardless of the orientation of the link it is attached to at start. Proof: if I configure a different initial orientation of the robot (in the .world file, e.g. initially bent or rotated), sensor measurements will still output (0, 0, 0) before moving any joint, even though we know that the orientation of the part the sensor is attached to is different. Increments in the orientation would be right, but the returned absolute values would not.

Even if we assumed that our robot would always initially start in the same pose, due to the orientation of the sensor frame (Z-down and pointing X-backwards, configured so to match the real robot), the measurements we are getting are reversed in sign. This should be doable via [.sdf configuration](http://sdformat.org/spec?ver=1.6&elem=sensor), but I didn't manage to get satisfactory results and it also didn't solve the main issue (forced alignment with the world frame).

traversaro:
Sorry (again) for the late reply. Thanks a lot for the clear description, this is quite subtle and problematic behaviour.  I think most of the users actually expects the IMU to return the ${}^{world} H_{sensor}$ (as enabled by your option), and not ${}^{sensorInitialPosition} H_{sensor}$, as that match what most real-world IMU actually use. So, even if it is a breaking change, do you think it could make sense to make the ${}^{world} H_{sensor}$ the default behaviour, and instead have an option to to keep the legacy behaviour, so that people that really need the ${}^{sensorInitialPosition} H_{sensor}$ can select it? I think we can name this opt-in flag `useInitialSensorOrientationAsReference` .

traversaro:
What do you think of what I proposed in the previous comment? @prashanthr05 @isorrentino @GiulioRomualdi @FabioBergonti @gabrielenava

isorrentino:
> What do you think of what I proposed in the previous comment?

I agree!

traversaro:
Cool @gabrielenava, I think that block is necessary nevertheless to handle the difference between the controller world and the IMU world, but anyhow the high level behaviour should be exactly the same before and after this modification.

Nicogene:
> What do you think of what I proposed in the previous comment? @prashanthr05 @isorrentino @GiulioRomualdi @FabioBergonti @gabrielenava @Nicogene

Ok also for me!

prashanthr05:
> What do you think of what I proposed in the previous comment? @prashanthr05 @isorrentino @GiulioRomualdi @FabioBergonti @gabrielenava @Nicogene

I agree with your proposed solution. It bypasses the cognitive overload of remembering to explicitly set the reference transformation. Also, this is a hard-to-debug problem, in my experience. So it is better to document this in detail.

traversaro:
@PeterBowman do you like to make the modifications to have the old behaviour as opt-in, or do you prefer that I do it? Thanks again!

PeterBowman:
Hi, @traversaro! I have add a config option for restoring the old behavior as opt-in, please feel free to further tweak it if you see fit.

Let me provide a few pics to better illustrate what was going on and how this change will impact the sensor's output.

default_gzclient_camera(1)-2023-03-08T20_09_44 760424

In the above image, our robot (TEO) is facing the X axis (Z is the vertical one). The IMU sensor is located in the chest. It is mounted upside-down: its local X axis should point in the opposite direction of global X, similarly the local Z is opposing global Z.

* Old orientation measurements: [0 0 0]
* New orientation measurements: [180 0 -180]

default_gzclient_camera(1)-2023-03-08T20_32_30 234390

Now, when we rotate the chest (with the IMU) +90 degrees wrt. global Z:

* Old orientation measurements: [0 0 -90]
* New orientation measurements: [180 0 -90]

default_gzclient_camera(1)-2023-03-08T20_11_00 704190

If TEO begins facing the Y axis instead (it has been rotated +90 degrees in global Z), but still in its initial pose:

* Old orientation measurements: [0 0 0]
* New orientation measurements: [180 0 -90]

The -X/-Z orientation issue (making the virtual sensor emulate how the real one is actually mounted) might/should be solved via proper .sdf configuration, but I tried several combinations of [`<orientation_reference_frame>` values](http://sdformat.org/spec?ver=1.6&elem=sensor#imu_orientation_reference_frame) and never managed to get it right. However, the real issue targetted here is the nonsense we get when TEO is oriented differently at start.

traversaro:
> However, the real issue targetted here is the nonsense we get when TEO is oriented differently at start.

Yes, that is the real "problem".

PeterBowman:
> Yes, that is the real "problem".

Hi! I'm terribly sorry, some meanings are difficult to catch via written word 😳. Is there some change I may have gotten wrong and should make to the current PR?

traversaro:
> > Yes, that is the real "problem".
> 
> Hi! I'm terribly sorry, some meanings are difficult to catch via written word 😳. Is there some change I may have gotten wrong and should make to the current PR?

No, sorry it was just a comment on my side, nothing preventing the merge of the PR.

PR #641, opened 2023-01-11
ControlBoardDriverCoupling: add couplings for hand mk5

Nicogene:
Hand Mk4 and mk5 only differ in the `thumb_rotate` that has been removed in the last revision of the hand (see https://icub-tech-iit.github.io/documentation/hands/hands_mk5/)

This has been tested with ergocub urdf


traversaro:
Do you have any reference/document describing the coupling laws of the hands? We can merge also without it, even if in that case it is not easy to review the equations.

plugins/controlboard/src/ControlBoardDriver.cpp
```diff
@@ -337,6 +337,12 @@ bool GazeboYarpControlBoardDriver::gazebo_init()
                 m_coupling_handler.push_back(cpl);
                 yCInfo(GAZEBOCONTROLBOARD) << "using icub_left_hand_mk4";
             }
+            else if (coupling_bottle->get(0).asString()=="hand_mk5")
```
traversaro, line 340: Can we add an `icub_` prefix here? It is not ideal to continue to put model-specific information here (see https://github.com/robotology/gazebo-yarp-plugins/issues/470), but it we do it let's at least make it clear in the public parameters to what robot/hands "family" we are referring to. Furthemore, for the `hand_mk4` the coupling were left specific, see https://github.com/robotology/gazebo-yarp-plugins/pull/620#issuecomment-1087364167 . Do you have any idea why instead for the mk5 this is not a problem?

pattacini:
> Do you have any reference/document describing the coupling laws of the hands? We can merge also without it, even if in that case it is not easy to review the equations.

Jot down a note for this in https://icub-tech-iit.github.io/documentation/hands/.

mfussi66 (approved):

plugins/controlboard/src/ControlBoardDriver.cpp
Nicogene: I changed the name in `icub_hand_mk5` and as far as I know, the mk4 coupling contained the `left` in the name because there was only a left hand for mk4, but trying it seems that that coupling is ok also for the right.

traversaro: For reference: we checked with @mfussi66 and he also confirmed that the coupling is not left or right specific, this was his assumption in https://github.com/robotology/gazebo-yarp-plugins/pull/620#issuecomment-1087364167 but it turned out that it was not true.

traversaro:
* apt is failing due to https://github.com/actions/runner-images/issues/6894
* conda/ubuntu due to https://github.com/robotology/idyntree-yarp-tools/pull/42

I fixed the conda failure, when the CI is happy can we merge and release v4.6.0 @Nicogene ?

Nicogene:
> * apt is failing due to [Unable to connect to azure.archive.ubuntu.com flaky failure on ubuntu-* strikes back actions/runner-images#6894](https://github.com/actions/runner-images/issues/6894)
> 
>     * conda/ubuntu due to [Fix conda-ci GitHub Action CI by using mambaforge directly idyntree-yarp-tools#42](https://github.com/robotology/idyntree-yarp-tools/pull/42)
> 
> 
> I fixed the conda failure, when the CI is happy can we merge and release v4.6.0 @Nicogene ?

Ok for me!

traversaro (approved):

Issue #643, opened 2023-01-13
CMake Error when linking `GazeboYARPPlugins` libraries


HosameldinMohamed:
When trying to build a plugin that uses some of the libraries of `gazebo-yarp-plugins`. Following http://robotology.github.io/gazebo-yarp-plugins/master/use_as_library.html

I get the following error when generating CMake:

```
 CMake Error in software/gazebo/plugins/vive-transform-plugin/CMakeLists.txt:
   Imported target "GazeboYARPPlugins::gazebo_yarp_lib_common" includes
   non-existent path

     "/home/hosameldin/code/robotology-superbuild/build/install/include/include"

   in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

   * The path was deleted, renamed, or moved to another location.

   * An install or uninstall procedure did not complete successfully.

   * The installation package was faulty and references files it does not
   provide.
```

However, the build goes through with no errors.

I thought it was worth reporting it anyways.

HosameldinMohamed:
CC @traversaro

traversaro:
Indeed, there is a bug in the CMake script. If you are compiling gazebo-yarp-plugins from source, can you check if https://github.com/robotology/gazebo-yarp-plugins/pull/644 solves the problem? Thanks!

traversaro:
> I thought it was worth reporting it anyways.

Thanks!

PR #644, opened 2023-01-16
Fix wrong install include for gazebo_yarp_lib_common library

traversaro:
Fix https://github.com/robotology/gazebo-yarp-plugins/issues/643

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ## [Unreleased]
 
+### Fixed 
+
+- Fix wrong install include for gazebo_yarp_lib_common library (https://github.com/robotology/gazebo-yarp-plugins/pull/644).
+
 ## [4.6.0] - 2023-01-13
 
 ### Added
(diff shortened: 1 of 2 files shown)


HosameldinMohamed (approved):
Thanks! I tested the installed library and the problem no longer exists!

I was searching for `INTERFACE_INCLUDE_DIRECTORIES` in the source code but couldn't find it!

HosameldinMohamed:
You may want to update the CHANGELOG too

traversaro:
> Thanks! I tested the installed library and the problem no longer exists!
> 
> I was searching for `INTERFACE_INCLUDE_DIRECTORIES` in the source code but couldn't find it!

That property is probably defined in the `GazeboYARPPluginsTargets.cmake` file that is generated and installed by CMake. That property is defined indirectly in the code via the `target_include_directories(gazebo_yarp_lib_common INTERFACE $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)` calls.

HosameldinMohamed:
> > Thanks! I tested the installed library and the problem no longer exists!
> > I was searching for `INTERFACE_INCLUDE_DIRECTORIES` in the source code but couldn't find it!
> 
> That property is probably defined in the `GazeboYARPPluginsTargets.cmake` file that is generated and installed by CMake. That property is defined indirectly in the code via the `target_include_directories(gazebo_yarp_lib_common INTERFACE $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)` calls.

Great! Thanks!

PR #653, opened 2023-05-29
Gazebo YARP Pedestrian SFM plugin

morpheus1820:
This is a plugin that enables simulation of crowds of people in Gazebo, with integrated obstacle detection, to be used in the CONVINCE project. The individual trajectories of simualted people can be modified via YARP rpcs.

It also contains an example gazebo world: (https://github.com/morpheus1820/gazebo-yarp-plugins/tree/sfm_plugin/plugins/pedestriansfm/worlds) 
and an example startup script: (https://github.com/morpheus1820/gazebo-yarp-plugins/blob/sfm_plugin/plugins/pedestriansfm/scripts/start_sim.sh)

P.S.:
The license of the original code from which this code is derived is BSD-3. I am not sure whether this is an issue for us.


traversaro:
Thanks @morpheus1820 for the contribution, can you sign the CLA? Thanks!

PR #665, opened 2023-09-17
ControlBoardDriverCoupling.cpp: include array header as array is used in the code

traversaro:
Fix https://github.com/robotology/robotology-superbuild/issues/1489 .


traversaro:
Let's merge to fix CI. fyi @Nicogene

PR #671, opened 2023-12-13
ControlBoardDriver: support use of IJointCoupling devices in gazebo_yarp_controlboard and drop support for hardcoded HandMK5 coupling

### LINKED ISSUE #470: Move gazebo_yarp_controlboard coupling logic to dynamically loaded plugins

Nicogene:
This PR wants to refactor the handMk5 coupling handler in order to use the `couplingXCubHandMK5` device:
- https://github.com/icub-tech-iit/ergocub-software/pull/178

And the `IJointCoupling` interface

- https://github.com/robotology/yarp/pull/3026

Moreover it simplify the handling of the coupling, the coupling hadler used to be a vector of handlers supporting more coupling per part, but since none is using this feature I changed the code in order to use only one coupling per part.

It fixes #470


Nicogene:
I tested these changes both with the robots that uses the "old" pipeline (e.g. iCubGazebo2_5_visuomanip) and ergoCubGazeboV1* and it works fine.

The only problem is that these changes requires YARP 3.9.0, then all the conda-based CI are broken, how can we proceed in this sense?

traversaro:
> The only problem is that these changes requires YARP 3.9.0, then all the conda-based CI are broken, how can we proceed in this sense @traversaro ?

I checked, the use of yarp 3.9 in stable branches is blocked by https://github.com/robotology/robotology-superbuild/pull/1548, basically the fact that icub-main master does not compile with yarp 3.9, and only icub-main devel compiles with yarp 3.9, so we can't use yarp 3.9 in stable branches of the superbuild at the moment. Can we quickly align you, me and @pattacini to understand how to handle this?

pattacini:
`icub-main@master` is now aligned with `icub-main@devel`.

pattacini:
> `icub-main@master` is now aligned with `icub-main@devel`.

Actually, this is not doable because of the icub-firmware-* deps.
`icub-main@master` reset to its previous commit.

@traversaro will apply a dedicated patch.

traversaro:
YARP 3.9 is now available with LatestReleases or Stable in the superbuild, and has been released on conda-forge:
* https://github.com/robotology/robotology-superbuild/pull/1548
* https://github.com/conda-forge/yarp-feedstock/pull/28

Nicogene:
> The only problem is that these changes requires YARP 3.9.0, then all the conda-based CI are broken, how can we proceed in this sense @traversaro ?

The CI is now fine!

plugins/controlboard/src/ControlBoardDriverInteractionMode.cpp
```diff
@@ -100,19 +100,35 @@ bool GazeboYarpControlBoardDriver::setInteractionMode(int j, yarp::dev::Interact
 {
     if (j < 0 || static_cast<size_t>(j) >= m_numberOfJoints) return false;
 
+    if (m_ijointcoupling) {
+        // TODO
```
traversaro, line 104:
```suggestion
```

Leftover?

traversaro (changes requested):
Minor comments, otherwise it seems ok. Can you write something on this in the CHANGELOG? Thanks!

traversaro:
cc @mebbaid

traversaro:
I think we can add "Fix https://github.com/robotology/gazebo-yarp-plugins/issues/470" in the PR body. :)

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format of this document is based on [Keep a Changelog](https://keepachangelo
 
 ## [Unreleased]
 
+### Changed
+
+- `controlBoard` plugin now uses a dinamically loaded plugin for handling the coupling of the hand mk5 (see https://github.com/icub-tech-iit/ergocub-software/pull/178)
```
traversaro, line 10: I think we should separate the two things:
* the fact that we add a new feature (loading the coupling from YARP device)
* the fact that we removed support for `COUPLING::icub_hand_mk5` parameters, and that if you want to still use ergocub models you need to update to load the pass the `device couplingXCubHandMk5` parameter.

The first point is interesting for all users of gyp, the second is a strictly related to ergocub users, my suggestion:

```suggestion
### Added
- The `gazebo_yarp_controlboard` plugin gained support to load at runtime arbitrary couplings specified by a yarp device that exposes the [`yarp::dev::IJointCoupling`](https://github.com/robotology/yarp/blob/v3.9.0/src/libYARP_dev/src/yarp/dev/IJointCoupling.h#L16) interface. The name of the yarp device used to load the coupling is passed via the `device` parameter in the `COUPLING` group. For an example of PR that uses this new feature, check https://github.com/icub-tech-iit/ergocub-software/pull/178 .

### Removed

- The support for passing the `icub_hand_mk5` parameter in the `COUPLING` the `gazebo_yarp_controlboard` plugin was removed. This breaks compatibility with models contained in ergocub-software <= 0.6.0, to use those models use a newer version of ergocub-software. 
```

traversaro: fyi @Nicogene

traversaro (approved):

traversaro:
@Nicogene can I merge with squash? Furthermore, are you ready to merge https://github.com/icub-tech-iit/ergocub-software/pull/178? I would merge (and eventually release, if we are ready to do so) both at the same time, to avoid breaking the superbuild Stable and Unstable branches.

Nicogene:
> @Nicogene can I merge with squash? 
Yes sure!
> Furthermore, are you ready to merge https://github.com/icub-tech-iit/ergocub-software/pull/178?
Yes I was awaiting to merge both PR togheter too

PR #677, opened 2024-04-02
Add support for loading vector from configuration files with parenthesis

traversaro:
As discussed in https://github.com/robotology/yarp/discussions/3092, it is recommended for future software that uses YARP to use vectors in configuration file by indicating them with parenthesis, i.e. to refer to them with:

~~~
nameOfList (elem1 elem2 elem3)
~~~

instead of:

~~~
nameOfList elem1 elem2 elem3
~~~

gazebo-yarp-plugins used to load the vectors from configuration files expecting no parenthesis. After this PR, now plugins support both the with parentheis and without parenthesis style, printing a warning if the without parenthesis style is used. If you can require gazebo-yarp-plugins 4.11.0, consider switching your configuration files to the vector with parenthesis style.

### DIFF
--- a/libraries/common/include/GazeboYarpPlugins/common.h
+++ b/libraries/common/include/GazeboYarpPlugins/common.h
@@ -9,6 +9,11 @@
 
 #include <string>
 #include <cmath>
+#include <vector>
+
+#include <yarp/os/LogStream.h>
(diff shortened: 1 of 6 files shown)


Nicogene (approved):
🚀

xela-95 (approved):

PR #681, opened 2024-04-08
gazebo_yarp_controlboard: Fix reading of actuatedAxesPosMax in COUPLING group that was broken in v4.11.0

Nicogene:
This PR fixes a regression introduced by the refactoring for handling the lists with parenthesis:
- #677 

~Draft because probably because some other problems I am having are under investigation~


traversaro (approved):

traversaro:
Thanks! Apparently the PR is not in draft, and for me it can be merged as it is (once CI passes), as for sure it fixes some problem. If there are more problems, we can fix them in subsequent PRs.

Nicogene:
> Thanks! Apparently the PR is not in draft, and for me it can be merged as it is (once CI passes), as for sure it fixes some problem. If there are more problems, we can fix them in subsequent PRs.

Yes I just put it ready for review because I found all the issues we can merge it thanks! 👍🏻

PR #684, opened 2024-08-21
ControlBoardDriver: Add check that min is lower than max in position limits

traversaro:
This may be useful to early detect possible typos in configuration files, that may be hard to debug otherwise.


traversaro:
fyi @kaidegast @mfussi66

mfussi66 (approved):

PR #696, opened 2025-02-25
Fix compilation with YARP master branch (future YARP 3.12) as of 2025/02/25

traversaro:
This PR switches migrate the Laser plugin to the use of yarp::dev::ReturnValue-based interface, as a follow up of:
* https://github.com/robotology/yarp/discussions/3168
* https://github.com/robotology/yarp/pull/3180
* https://github.com/robotology/yarp-device-rplidar/pull/7

To avoid the need to keep two branches, one for compatibility with YARP 3.11.* and one for the future YARP 3.12.*, I prepared a `YarpDevReturnValueCompat.h` header that can be useful to implement an interface that works in both cases. As these header can be useful also for other repos, but I do not think it make sense to install in an a reusable library as it is a temporary solution, I will store it in https://gist.github.com/traversaro/ab7ab377c70c50d312a7b5edd4da6242, so that interested users can copy it in their repo as a private (non-installed) header to simplify the support of the future YARP 3.12

Similar to https://github.com/robotology/gz-sim-yarp-plugins/pull/240 .

### DIFF
--- a/libraries/common/include/GazeboYarpPlugins/YarpDevReturnValueCompat.h
+++ b/libraries/common/include/GazeboYarpPlugins/YarpDevReturnValueCompat.h
@@ -1,0 +1,44 @@
+#ifndef YARP_DEV_RETURN_VALUE_COMPAT_H
+#define YARP_DEV_RETURN_VALUE_COMPAT_H
+
+// Defines YARP_VERSION_* macro
+#include <yarp/conf/version.h>
+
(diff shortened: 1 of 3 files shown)


Nicogene (approved):

traversaro:
> @pattacini a PR like this one would be necessary also in icub-main at some point

Yes, feel free to reuse the compat header if you find it useful.

